PR #2, opened 2016-11-29
Update sample specification template

cybernth:
Added impacted modules definition in the sample specification


cybernth:
@r0h4n  @brainfunked  please review

r0h4n (approved):

PR #6, opened 2016-11-30
Added specification for refactoring gluster get-state output

shtripat:
Current output of get-state command from gluster also provides
volume options details. This specification talks about changes
required for accomodating the same and updating details in Tendrl
central store.

### DIFF
--- a/specs/refactor_gluster_integration_get_state_dump.adoc
+++ b/specs/refactor_gluster_integration_get_state_dump.adoc
@@ -540,7 +540,7 @@ None
 
 == Implementation
 
-* github issue link here
+* https://github.com/Tendrl/specifications/issues/30
 
 === Assignee(s)
 
@@ -582,3 +582,5 @@ None
 == References
 
 * https://www.redhat.com/archives/tendrl-devel/2016-November/msg00063.html
+
+* https://github.com/Tendrl/gluster_integration/issues/41
--- a/specs/refactor_gluster_integration_get_state_dump.adoc
+++ b/specs/refactor_gluster_integration_get_state_dump.adoc
@@ -148,15 +148,344 @@ Last allocated port: 49153
 
 * General refactoring of code handling the gluster get-state output and add
 all missing volume attributes to the Volume object
-** Additional fields available under volume details from get-state output
-like transport_type, snap_count, stripe_count, replica_count, arbiter_count,
-quorum_status, ... to be added to central store
+** Below additional fields available under volume details from get-state output
+to be added to central store
+```
+- transport_type
+- snap_count
+- stripe_count
+- replica_count
+- subvol_count
+- arbiter_count
+- disperse_count
+- redundancy_count
+- quorum_status
+- snapd_svc.online_status
+- snapd_svc.inited
+- rebalance.id
+- rebalance.status
+- rebalance.failures
+- rebalance.skipped
+- rebalance.lookedup
+- rebalance.files
+- rebalance.data
+```
 
 * Modify the tendrl definitions specifications for gluster and add all the
 missing attributes to Volume object
+** The snippet from updated tendrl definitions would be as below for volume
+attributes
+```
+     Volume:
+       .......
+       attrs:
+         arbiter_count:
+           help: "Arbiter count of volume"
+           type: Integer
+         bricks:
+           help: "List of brick mnt_paths for volume"
+           type: List
+         deleted:
+           help: "Flag is volume is deleted"
+           type: Boolean
+         disperse_count:
+           help: "Disperse count of volume"
+           type: Integer
+         disperse_data_count:
+           help: "Disperse data count of volume"
+           type: Integer
+         force:
+           help: "If force execute the action"
+           type: Boolean
+         redundancy_count:
+           help: "Redundancy count of volume"
+           type: Integer
+         replica_count:
+           help: "Replica count of volume"
+           type: Integer
+         stripe_count:
+           help: "Stripe count of volume"
+           type: Integer
+         transport:
+           help: "Transport type for volume"
+           type: String
+         vol_id:
+           help: "ID of the gluster volume"
+           type: String
+         volname:
+           help: "Name of gluster volume"
+           type: String
+         vol_type:
+           help: "Type of the volume"
+           type: String
+         rebal_data:
+           help: "Rebalance data"
+           type: String
+         rebal_skipped:
+           help: "Skipped files while rebalance"
+           type: String
+         subvol_count:
+           help: "Count of subvolumes"
+           type: Integer
+         brick_count:
+           help: "Count of bricks"
+           type: Integer
+         cluster_id:
+           help: "UUID of the cluster"
+           type: String
+         snapd_inited:
+           help: "If snapd is initialized"
+           type: String
+         status:
+           help: "Status of the volume"
+           type: String
+         rebal_id:
+           help: "UUID of the rabalance task"
+           typoe: String
+         rebal_lookedup:
+           help: "Looked up files for rebalance"
+           type: Integer
+         snap_count:
+           help: "Count of the snapshots"
+           type: Integer
+         rebal_files:
+           help: "No of files rebalanced"
+           type: Integer
+         snapd_status:
+           help: "Status of snapd"
+           type: String
+         options:
+           help: "options list for volume"
+           type: json
+         quorum_status:
+           help: "Quorum status"
+           type: String
+         rebal_status:
+           help: "Status of rebalance task"
+           type: String
+         rebal_failures:
+           help: "Failed no of files for rebalance"
+           type: Integer
+       enabled: true
+       ..........
+```
 
 * Add an attribute namely "options" to Volume object which is of type json and
 contains key:value pairs as option name and option values
+
+** The new structure of volume object under etcd would be as below after changes
+```
+{
+
+    "action": "get",
+    "node": {
+        "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535",
+        "dir": true,
+        "nodes": [
+            {
+                "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535/vol_type",
+                "value": "Distribute",
+                "modifiedIndex": 4647,
+                "createdIndex": 4647
+            },
+            {
+                "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535/rebal_data",
+                "value": "0Bytes",
+                "modifiedIndex": 4630,
+                "createdIndex": 4630
+            },
+            {
+                "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535/rebal_skipped",
+                "value": "0",
+                "modifiedIndex": 4635,
+                "createdIndex": 4635
+            },
+            {
+                "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535/subvol_count",
+                "value": "1",
+                "modifiedIndex": 4644,
+                "createdIndex": 4644
+            },
+            {
+                "key": "/clusters/7aece92f-7bae-47eb-9511-0e68b0664e5e/Volumes/ff0bceae-9901-46b0-9464-7fa4836d2535/Bricks",
+                "dir": true,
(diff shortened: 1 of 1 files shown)


shtripat:
@r0h4n @brainfunked @nthomas-redhat @anmolbabu @nnDarshan please review.

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,235 @@
+= Refactor Gluster Integration's state dump module
+
+
+The gluster get-state output is consumed by tendrl-gluster-integration. We need
+to re-factor the way Tendrl consumes and processes the get-state output.
+
+Notes:
+
+* In light of new changes to gluster get-state
+  http://review.gluster.org/#/c/15889
```
r0h4n: Update new review http://review.gluster.org/#/c/15975

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,235 @@
...
+Primary assignee:
+  shtripat
+
+
+Other contributor(s):
+  r0h4n
+
+== Work items
+
+* Introduce a class VolumeOptions in persister module to render volume options
```
r0h4n, line 556: Volume options should be part of the same Volume class, No new class is required, please update

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,235 @@
...
+Peer2.connected: Connected
+Peer2.othernames:
+
+[Volumes]
+Volume1.name: vol1
+Volume1.id: 5e24f6ed-239e-4961-ac47-1e9ed44aa7b1
+Volume1.type: Distribute
+Volume1.transport_type: tcp
+Volume1.status: Started
+Volume1.brickcount: 3
```
r0h4n, line 61: Please ensure all these new attributes and options are reflected in the "Volume" object itself

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+== Dependencies
+
+* Required Gluster 3.9 + patch http://review.gluster.org/#/c/15889
+
+== Testing
+
+* Check if the gluster get-state output is parsed without any error
+
+* Verify the volume details in central store if it contains the volume options
```
r0h4n, line 572: Lets not differentiate betweeen volume attributes and options, they are same and should go into "Volume" object

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+* Introduce a class VolumeOptions in persister module to render volume options
+in central store (etcd)
+
+* Update the get-state output parsing logic and update volume options details to
+central store
+
+== Dependencies
+
+* Required Gluster 3.9 + patch http://review.gluster.org/#/c/15889
```
r0h4n: Update new review http://review.gluster.org/#/c/15975

r0h4n (changes requested):

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done. Will add all other attributes of volume.

But volume options names are not statically defined for a volume. The list gets changed based on what features (e.g. snapshot, tiering, quota, aribter, nfs-ganesha) are enabled for the volume. Rather "Options" need to be an attribute for the volume under which a dictionary represents the various options set for the volume.

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: As commented in previous comment

"The volume options names are not statically defined for a volume. The list gets changed based on what features (e.g. snapshot, tiering, quota, aribter, nfs-ganesha) are enabled for the volume. Rather "Options" need to be an attribute for the volume under which a dictionary represents the various options set for the volume."

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: As  commented earlier 

"The volume options names are not statically defined for a volume. The list gets changed based on what features (e.g. snapshot, tiering, quota, aribter, nfs-ganesha) are enabled for the volume. Rather "Options" need to be an attribute for the volume under which a dictionary represents the various options set for the volume."

We actually need to differentiate between normal volume attributes like "id", "name" etc from options. Rather we should have an attribute called "options" and a dict of different available options need to be listed under this attribute.

specs/refactor_gluster_integration_get_state_dump.adoc
r0h4n: It is ok if the option visibility changes based on features. 

We still need to check for each option and define them in Tendrl definitions and treat them as attributes.

specs/refactor_gluster_integration_get_state_dump.adoc
r0h4n: Why do we need to differentiate?

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Based on features enabled or disabled these options may appear/disappear from the list. Not sure why we need to keep static in tendrl only whereas its actually dynamic in gluster CLI output.

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: We need to differentiati because they are not volume attributes. Rather "options" is volume attribute.
Also if volume features are enabled dynamically runtime, we dont need to change anything in tendrl for this and autimatically all the keys would be available in tendrl as part of refresh of cluster data.

specs/refactor_gluster_integration_get_state_dump.adoc
r0h4n: You need to define all these options/attributes via Tendrl definitions because tendrl-api, tendrl-dashboard, tendrl-gluster-integration can discover these. 

If not defined in Tendrl definitions, we must not use them in our code. This is clear violation of our definitions driven approach

@brainfunked @nthomas-redhat @anivargi ^

r0h4n: You need to define all these options/attributes via Tendrl definitions because tendrl-api, tendrl-dashboard, tendrl-gluster-integration can discover these. 

If not defined in Tendrl definitions, we must not use them in our code. This is clear violation of our definitions driven approach

@brainfunked @nthomas-redhat @anivargi ^

shtripat: As long as I understand we should define one attribute "options" for the volume object in definitions which should be a dictionary. If dictionary is not supported we need to look at that and see how that could be supported.

This attribute "options" could be involved in below scenarios -
1. Create Volume: Volume volume create only default options would be set by gluster and post creation when volume gets refreshed in tendrl, it would get the dict with default keys in "options" dictionary.
2. Edit Volume: During edit volume there are two options. Either change the value of the existing option or add new one. Adding new one is more driven by enabling a new feature for volume rather than manually adding the key and value in "options"
3. Listing of volume: The api needs to get the "options" as well, the way they get the bricks and append to the output as a dict

I feel, as long as definitions support dictionary type it should be achievable and more generic in nature. Otherwise in future when gluster adds another feature, we need to update tendrl code again and again only for this reason.

@nthomas-redhat @brainfunked comments??

r0h4n: Great, this gives more clarity

we can work with an "options" dict/map in definitions only if:
- All the options have a default values (None or whatever) set by gluster itself on activation of specific features


Impact:
Since we need to process "dict" type via our definitions file, this should be part of this spec or atleast create a new spec and make this spec depend on that.

shtripat: So, its like if a feature for the volume is enabled, yes all the required volume options fir sure would be set with some default values as long as I remember :)

I feel a new spec for handling "dict" type in definitions is fine and linking with this. But do we block this spec from merging or we can go ahead and handle "dict" support in definitions as separate item?

r0h4n: Both specs can continue in parallel, But you would need the "dict" type changes in place to test your spec.

Please document the other spec as well and link it to your spec.

shtripat: As discussed volume options attribute could be a json which contains key:value pairs.
Will update this spec with details of changes to be done in definitions as well.

r0h4n (approved):

r0h4n:
@nthomas-redhat @brainfunked plz review

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,252 @@
...
+
+== Problem description
+
+Current Tendrl way of consuming and processing the gluster get-state output
+considers only the volumes and bricks and pushes the details to central store.
+It needs to be modified to update the newly available volume options as well to
+central store.
+
+Also not all the volume attributes are added to Volume object in central store.
+The same should be taken care and added.
```
brainfunked: Should be: "The same should be taken care of and added."

specs/refactor_gluster_integration_get_state_dump.adoc
brainfunked: Regarding the Edit Volume scenario mentioned above by @shtripat, we need to document which specific scenarios will need to be handled as individual flows, and which can be satisfied directly by a generic UpdateVolumeOptions or similar. This has impact on the definitions and needs to be documented in this spec.

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -0,0 +1,252 @@
...
+svc5.name: quotad
+svc5.online_status: Offline
+
+
+[Misc]
+Base port: 49152
+Last allocated port: 49153
+```
+
+* General refactoring of code handling the gluster get-state output and add
```
brainfunked, line 149: The details of the refactoring need to be specified below.

brainfunked (changes requested):

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: As long as gluster is concerned these options are set through features enable/disable and also can be set/edited individually using CLI.

What I have thought for tendrl is that, provide a edit option for the volume and let the user edit any option which is already set using a single flow say named as UpdateVolumeOptions.

These volume options can be set/edited as well from different work-flows in tendrl UI as well. 

Why I suggest this way because some of the options get set through volume creation options whereas some of them are set as specific volume set option CLI.

specs/refactor_gluster_integration_get_state_dump.adoc
```diff
@@ -148,6 +148,9 @@ Last allocated port: 49153
 
 * General refactoring of code handling the gluster get-state output and add
 all missing volume attributes to the Volume object
+** Additional fields available under volume details from get-state output
+like transport_type, snap_count, stripe_count, replica_count, arbiter_count,
+quorum_status, ... to be added to central store
 
 * Modify the tendrl definitions specifications for gluster and add all the
 missing attributes to Volume object
```
cybernth, line 175: - list down all attributes you are planning to add
- structure json which will updated to etcd
- Any changes to tendrl definitions to be written down here(snip from tendrl definitions)

cybernth (changes requested):

specs/refactor_gluster_integration_get_state_dump.adoc
shtripat: Done

cybernth:
@shtripat , Looks good to me. Only thing I found is the github issue is missing in the spec. You need to file an issue on specifications repo. Also file issues on all the components where the chnages are made. Refer these issues in the refereces section of the documnent.

shtripat:
@nthomas-redhat added issues in references section. please review..

cybernth (approved):
LGTM

cybernth:
https://github.com/Tendrl/specifications/issues/30

PR #7, opened 2016-12-01
Added specifications for refactoring tendrl code

shtripat:
This specification details the refactoring across the tendrl
modules to move major common code to common module.

### DIFF
--- a/specs/refactor_common_code_and_move.adoc
+++ b/specs/refactor_common_code_and_move.adoc
@@ -56,6 +56,10 @@ the same in individual modules
 * Re-factor the code across various modules to fix test cases and attain the
 code coverage level to 90%
 
+* Move the command executor, package installer and service manager utilities
+from node_agent module to common module as its generic in nature and can be
+used by other modules as well.
+
 
 === Alternatives
 
@@ -82,6 +86,7 @@ None
 ** Base atom class
 ** Base test case class
 ** Configuration management logic
+** Command executor, package installer and service manager utilities
 
 ==== Tendrl/node_agent impact
 
@@ -96,6 +101,9 @@ None
 * Extend all the atoms from common atom class
 
 * Extend all the test cases from common test class
+
+* Remove command executor, package installer and service manager as its
+now moved to common module
 
 ==== SDS integration impact
 
@@ -171,6 +179,10 @@ to different modules as long as they provide the location details of these
 configuration files. The common module as library doesn't default to any
 location for these configuration files.
 
+* The node agent flows which invoke command executor, package installer and
+service manager now need to refer common module utilities as the code is moved
+to common module now.
+
 === Assignee(s)
 
 Primary assignee:
@@ -178,7 +190,7 @@ Primary assignee:
 
 
 Other contributor(s):
-  r0h4n
+  r0h4n, nnDarshan
 
 == Work items
 
@@ -212,6 +224,9 @@ module
 
 * Re-factor the test cases and make sure code coverage is 90% or more
 
+* Move the command executor, package installer and service manager utilities
+to common module
+
 
 == Dependencies
 
@@ -228,7 +243,10 @@ None
 
 == Documentation impact
 
-None
+* Update the tendrl/common document and remove the logging yaml related
+changes as the common module is now used as a library and doesnt have
+its own logging. Rather it would use specific module's logging e.g.
+node-agent, gluster-integration etc.
 
 == References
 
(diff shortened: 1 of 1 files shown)


shtripat:
@nthomas-redhat @r0h4n @brainfunked @anmolbabu @nnDarshan @GowthamShanmugam please review.

r0h4n (changes requested):
Separate out refactoring of each component into individual specs, common, tendrl-gluster-integraton etc

shtripat:
@r0h4n I was of a thought that each module specific code changes would be created as individual github issues and this spec PR would be linked to all of them. So effectively this spec PR ends up into multiple github issues and multiple re-factoring code change PRs as a result.

@nthomas-redhat @brainfunked comments?? ^^

If everybody strongly feels, its required, I can create multiple ones (its just a copy paste of same content mostly ;))

specs/refactor_common_code_and_move.adoc
```diff
@@ -0,0 +1,206 @@
+= Refactor modules like tendrl-gluster-integartion, tendrl-node-agent etc and
```
brainfunked: Line breaks do not work in titles, if I'm not mistaken. Shorten the main title. Add a subtitle if you wish to convey additional context.

specs/refactor_common_code_and_move.adoc
```diff
@@ -0,0 +1,206 @@
...
+business logic.
+
+Even there are various sample configuration files which are replicated across
+the tendrl modules and could be more organized and moved to common module. Only
+modules specific configurations should be kept as part of the individual modules
+
+
+== Use Cases
+
+None
```
brainfunked: The most important use case is that this enables commons to function as a library and the code becomes completely reusable by any component.

brainfunked (changes requested):
Need implementation details regarding which specific flows need to be moved.

brainfunked:
The following issues need to be incorporated in this specification:
* https://github.com/Tendrl/common/issues/58
* https://github.com/Tendrl/common/issues/60
* https://github.com/Tendrl/common/issues/62

r0h4n:
@shtripat Specs needs to be per component and github issues should be against that component, It is difficult to track one spec spanning multiple components. Also that way component owners can review those specs without needing much insight on other components apart from making sure the impact on other components is documented

brainfunked:
@r0h4n Specifications are tendrl-global. They can affect multiple components. The specification should explicitly point out the affected components and the changes required across them. Issues from the specification can be created across multiple components. The issues detail out the exact implementation changes required in each of the components.

Basically, specifications are our tendrl project design. Issues are per component implementation details.

specs/refactor_common_code_and_move.adoc
shtripat: Ok. Will shorten the heading..

specs/refactor_common_code_and_move.adoc
shtripat: Done. Would add the use case details.

shtripat:
@brainfunked Agree. Even I feel this specification could be single and we can add all the required details here and then link the github issues to this from different modules of tendrl.
I would prefer retaining this as single spec.

nnDarshan:
@shtripat Can you please mention about moving the general purpose utilities like command executor, package installer, service manager etc  to common in this spec.
I see you have already mentioned about utils, but they are more specific to flows, manager etc.

shtripat:
@nnDarshan regarding your comment, I think enough details are available under "Proposed changes" and "Work Items". These section talk about manager, persister changes etc in detail. Anything specific you need?

nnDarshan:
@shtripat Can you please mention about general purpose utilities like command executor, package installer, service manager.

specs/refactor_common_code_and_move.adoc
```diff
@@ -203,4 +232,8 @@ None
 
 == References
 
-None
+* https://github.com/Tendrl/common/issues/58
+
+* https://github.com/Tendrl/common/issues/60
+
+* https://github.com/Tendrl/common/issues/62
```
cybernth, line 257: Have a mention about issue 60 as well in the document. That means common module specific logging and configuirations will be removed. This has a documentaion impact as well

Also you need to create individual issues per component(changes occuring) and link it here

cybernth (changes requested):
overall LGTM if the comments are addressed

specs/refactor_common_code_and_move.adoc
shtripat: Done

nnDarshan: @shtripat can you please include: https://github.com/Tendrl/common/issues/80. Sorry for informing you late.

shtripat: Done

cybernth (approved):
Looks good to me

r0h4n (approved):

PR #8, opened 2016-12-01
Add spec for flow framework refactoring

anmolbabu:
Add spec for flow framework refactoring


anmolbabu:
@shtripat @nthomas-redhat @r0h4n @brainfunked Please review

r0h4n (changes requested):
spec name "monitoring_defs" is too vague.

specs/flow_defs_reorganization.adoc
```diff
@@ -0,0 +1,287 @@
...
+
+== Use Cases
+
+* Monitoring code cannot be part of any component of tendrl core stack.
+  The monitoring stack needs to provide a flow for configuring the monitoring
+  agents maintained by tendrl.
+
+== Proposed change
+
+* The node-agent pushes a master definition /tendrl_definitions to etcd which
```
brainfunked: We should have a namespace `tendrl` in etcd, under which no directory or file needs to have the `tendrl` prefix.

specs/flow_defs_reorganization.adoc
```diff
@@ -0,0 +1,287 @@
...
+  node_agent reads the master definitions in /tendrl_definitions in etcd
+  and subsequently reads all definitions in /tendrl_integrations.d/ on
+  seeing the line "include /tendrl_integrations.d/*".
+
+
+=== Alternatives
+
+* Append monitoring related definitions yaml to node_agents definitions
+  directory in etcd.
+
```
brainfunked: Please fix the formatting here. Don't think blank lines are allowed between list items.

brainfunked (changes requested):
This specification also requires three other specifications:
* To include additional definition files via an `include` statement, that should be valid only in the master definition file for now.
* Concept of reserved namespaces and namespace overrides. Definition files loaded via include statements should be able to override existing namespaces and objects. However, some namespaces, for instance the global tendrl namespace, should be reserved and not be allowed to be overridden. These are the reserved namespaces.
* Compilation of the definition files together should be done via a definition refresh flow in the node agent. The compiled, merged namespaced representation of all the definition files combined is what every component should refer to.

specs/flow_defs_reorganization.adoc
anmolbabu: Done

specs/flow_defs_reorganization.adoc
anmolbabu: Done

specs/monitoring_defs.adoc
```diff
@@ -0,0 +1,105 @@
...
+  to create the definitions path instead of hard-coding it.
+* And from there the already existing flow framework as usual will extract and
+  execute the pre, post and the normal flow atoms.
+
+=== Alternatives
+
+* Append monitoring related definitions yaml to node_agents definitions
+  directory in etcd.
+
+    ** Drawbacks
```
shtripat: I dont think you need tabs if you are using ```**``` and ```***```. They automatically would happen in final doc

shtripat (commented):
I dont see any major comments as such. Looks fine to me as long as other comments are taken care.

anmolbabu:
@brainfunked 
'To include additional definition files via an include statement, that should be valid only in the master definition file for now.
Concept of reserved namespaces and namespace overrides. Definition files loaded via include statements should be able to override existing namespaces and objects. However, some namespaces, for instance the global tendrl namespace, should be reserved and not be allowed to be overridden. These are the reserved namespaces.'
These are to be done as part of flow validation right?
@r0h4n Could you please help raising specs for these
'Compilation of the definition files together should be done via a definition refresh flow in the node agent. The compiled, merged namespaced representation of all the definition files combined is what every component should refer to.'
@r0h4n is this handled as pert of any of your PRs already? Probably as part of spec for:
'Every component will build an in-memory map of the definitions based on the namespace versions in etcd. A namepsace can be assembled using multiple yaml files. Every part of the namespace (objects, flows etc.) would be referenced based on the latest version available for it.'

r0h4n:
Ill work on this one @anmolbabu , you can focus on the monitoring flows

anmolbabu:
Closing this as @r0h4n has addressed it.

PR #11, opened 2016-12-01
Specification: Separate out per-component configuration files

GowthamShanmugam:

### DIFF
--- a/specs/separate_configuration_files_per_component.adoc
+++ b/specs/separate_configuration_files_per_component.adoc
@@ -1,0 +1,234 @@
+= Specification for segregating configurations per module wise
+
+This specification describes, Currently all the configuration details of
+different modules are read from same file. Instead of that store the
+configuration details in different files as well as etcd and read from
+particular source based on priority.
+
+== Problem description
+
+Currently, All the tendrl modules are configured from same file. Also, while
+import flow the configuration file gets appended with SDS specific
+configurations while installation of SDS integration bits on the storage nodes.
+To avoid this we have to store the configuration details into layer wise.
+
+1. Store the configuration details into different files as per modules, And
+   configure modules using those files. (high priority).
+2. Each modules should store its own configuration details to etcd and configured
+   via api request (Less priority).
+
+Only node_agent configuration file have its own configuration details and
+etcd configuration details.
+
+At the time of installation, Each module's configuration details should store in
+different files with proper name. Always tendrl modules should configured from
+those files only.
+When module run the configuration details should be stored in etcd.
+
+* Integration modules should store its configuration details under cluster
+  specific hierarchy in the central store.
+* Other modules should store its configuration with proper hierarchy.
+
+The configuration in the central store needs to be treated as an object
+hierarchy, which appropriate atoms and flows to allow modification of this
+object hierarchy. If any modification on configurations in etcd, Then it should
+affect a local configuration files also. But any changes in local configuration
+file should not affect configurations in etcd. Any changes in integration
+specific configuration in etcd will affect config files of particular cluster
+nodes only. But changes in other configuration in etc will affect all nodes
+local config files.
+
+== Use Cases
+
+Centralized deployment-wide or cluster-wide configuration in the central store,
+which can be overridden per component via configuration files or command line
+options.
+
+== Proposed change
+
+* Modify the sample configuration files in each modules. It should contain
+  its own and relevant configuration information only.
+* Rename the sample configuration files in each modules with proper name.
+* Modify the spec file in each module to copy the configuration file with
+  proper name.
+* Create a new python file each module to keep configuration details.
+* Modify the config file in each module  to read the configuration details from
+  particular local config file.
+* Modify the manager in each module to read the configuration details from newly
+  created file and store it in in etcd. Integration specific modules should
+  store configuration details under cluster specific hierarchy in the central
+  store.
+* Modify the yaml file in each module with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Create atoms and flows to allow modification on configuration in etcd. It
+  should modify local configuration file also.
+* Modify tendrl definition file in node_agent to store node_agent specific
+  configuration details.
+
+== Alternatives
+
+None
+
+== Data model impact
+
+* Define and implement a "Config" object for each Tendrl component
+* Add config details in etcd with key /_tendrl/Config/repo_name.
+* Add integration specific details in etcd with key
+  /clusters/cluster_id/_tendrl/Config/repo_name.
+
+== Impacted modules
+
+=== Tendrl API impact:
+
+* Api to get deployment wise configuration details /_tendrl/Config/repo_name.
+* Api to get integration specific configuration details
+  /clusters/cluster_id/_tendrl/Config/repo_name.
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+* Modify the tendrl definition file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd with proper
+  key.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+=== Tendrl/gluster_integration impact
+
+* Modify the installation document to copy configuration file with proper name.
+* Rename sample configuration file
+* Modify the spec file to copy configuration file with proper name.
+* Modify the tendrl definition file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd under cluster
+  specific hierarchy in the central store.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+=== Tendrl/ceph_integration impact
+
+* Modify the installation document to copy configuration file with proper name.
+* Rename sample configuration file
+* Modify the spec file to copy configuration file with proper name.
+* Modify the tendrl definition file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd under cluster
+  specific hierarchy in the central store.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+=== Tendrl/performance_monitoring impact
+
+* Modify the installation document to copy configuration file with proper name.
+* Rename sample configuration file
+* Modify the spec file to copy configuration file with proper name.
+* Modify the yaml file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd with proper
+  key.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+=== Tendrl/alerting impact
+
+* Modify the installation document to copy configuration file with proper name.
+* Rename sample configuration file
+* Modify the spec file to copy configuration file with proper name.
+* Modify the yaml file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd with proper
+  key.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+None
+
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
+* https://github.com/Tendrl/gluster_integration/issues/81
+* https://github.com/Tendrl/ceph_integration/issues/54
+* https://github.com/Tendrl/node_agent/issues/91
+* https://github.com/Tendrl/common/issues/76
+* https://github.com/Tendrl/performance_monitoring/issues/11
+* https://github.com/Tendrl/alerting/issues/7
+
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* Modify the installation document to copy configuration file with proper name.
+* Rename sample configuration file
+* Modify the spec file to copy configuration file with proper name.
+* Modify the yaml file with new object, atoms, flows, pre-run,
+  post-run, attributes, type, uuid, version, inputs, description for the
+  configuration.
+* Add new python file to keep configuration details.
+* Modify manager to read and store it own configuration into etcd with proper
+  key.
+* Create new flow and atoms to allow modification on configuration in etcd. And
+  it should affect local configuration file also.
+
+== Dependencies
+
+* https://github.com/Tendrl/specifications/pull/7
+
+== Testing
+
+* Sanity check for flow
+* Check all modules are configured successfully
+* Check all the logging are works fine
+
+== Documentation impact
+
+* Modify installation documents in ceph_integration, gluster_integration,
+  node_agent, alerting, performance_monitoring.
+
+== References
+
+* https://github.com/Tendrl/node_agent/issues/83
+* https://github.com/Tendrl/node_agent/issues/99
+* https://github.com/Tendrl/gluster_integration/issues/86
+* https://github.com/Tendrl/ceph_integration/issues/57


specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
+= Separate out per-component configuration files
```
shtripat: Change the heading as "Specification for segregating configurations per module wise"

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+
+Configuration details for the tendrl modules are read from same file(tendrl.conf).
+Based on the tendrl modules we have to keep the configuration details into different
+files with proper name and read it independently.
+
+== Problem description
+
+Currently all the configuration details for different modules are read from same file .
```
shtripat: Also while import flow the configuration file gets appended with SDS specific configurations while installation of SDS integration bits on the storage nodes. This also should be mentioned in problem statement.

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+We have to split the configuration details as per modules and keep that into different
+file with proper name. Each configuration file should have its own and relevant configuration
+information only.
+
+We have to modify the python file (config.py) which read configuration file as per this
```
shtripat: This is rather implementation detail and not problem statement

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+modification. The python file should read each configuration files and it has to combine all
+the information into single object.
+
+== Use Cases
+
+None
+
+== Proposed change
+
+** Modify the sample configuration files in each modules. It should contains its own
```
shtripat: %s/should contains/should contain/g

shtripat: Why second level bullet?

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+   the name of conf.d.
+** Modify the spec file in each modules to copy the sample configuration file into proper
+   directory:
+      * From common repository modify the spec file to copy the sample configuration file
+        inside the directory /etc/tendrl.
+      * From gluster_integration, ceph_integration and node_agent modify the each spec files
+        to copy the sample configuration files inside the directory /etc/tendrl/conf.d.
+** Modify config.py from common repository to read each configuration files independently
+   and it should combine all the information into single object.
+** From node_agent repository modify copy statement in /doc/source/installation.rst
```
shtripat: The below three points should go inside document impact section I feel

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+** etc/tendrl/tendrl.conf.sample
+** config.py
+** tendrl-ceph-integration.spec
+
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
```
shtripat: Even notifications and monitoring modules would contain config files and should be taken care

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+  node_agent, common.
+* Change config.py to read all configuration file in common.
+* Change installation document in gluster_integration, ceph_integration,
+  node_agent, common.
+* Change spec file in gluster_integration, ceph_integration,
+  node_agent, common.
+
+== Dependencies
+
+* It is dependent on refactor job.
```
shtripat: Put a link to the PR here

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+
+== Testing
+
+* Check all configuration files are read successfully
```
shtripat: Also a sanity of one major flow should be performed as part of testing this I feel. Also make sure all the logging etc is happening properly.

specs/Separate_conf_files.adoc
```diff
@@ -0,0 +1,141 @@
...
+
+== Documentation impact
+
+* All installation documents needs some minior change
```
shtripat: Provide the exact details of the changes required in documents here rather

r0h4n (changes requested):
specifications names should be unique and they should be in snakecase.

rename to "separate_configuration_files_per_component.adoc"

brainfunked:
I think that it would be a good idea to update this specification to allow configuration to be centralised via etcd, as discussed at https://github.com/Tendrl/alerting/pull/1#issuecomment-264384626. On the systems, `tendrl.conf` should be shipped via `commons` and utilised from thereir to connect to etcd. So, this leads to changes in `commons` for etcd connection handling using this configuration file.

@r0h4n @shtripat @nthomas-redhat

specs/Separate_conf_files.adoc
GowthamShanmugam: i will rename it

specs/Separate_conf_files.adoc
GowthamShanmugam: i have added some sub bullets points under this bullet points, So it will separate bullet points from sub bullet points

GowthamShanmugam: i will change it

specs/Separate_conf_files.adoc
GowthamShanmugam: i will remove this points, i have already added this points in implementation section. These are duplicates so not required.

specs/Separate_conf_files.adoc
GowthamShanmugam: i will move this points to documentation impact part

specs/Separate_conf_files.adoc
GowthamShanmugam: i will add it

GowthamShanmugam:
@brainfunked  Why we have to store config details in etcd, Can you please explain the process clearly.

r0h4n:
@brainfunked With configuration for anything not common residing in etcd, that would require users to modify configuration via etcd. Do we make the users manually edit the configuration in etcd?

GowthamShanmugam:
@shtripat , @anmolbabu , @r0h4n @brainfunked please review it

mbukatov:
Aah, when I see term *specfile* again, I would not know outright if that means ascii doc formatted file with specification description from Tendrl/specifications, or if it refers to some rpm specfile (`*.spec`). I guess I would need to start using *rpm specfile* to be clear. On the other hand, we may confuse newcomers this way.

GowthamShanmugam:
@mbukatov  Oh, it is my mistake will rename it to specification. And thanks for point out this issue :)

GowthamShanmugam:
@r0h4n @shtripat @anmolbabu please review it

GowthamShanmugam:
@shtripat @brainfunked @r0h4n please review it

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,157 @@
+= Specification for segregating configurations per module wise
+
+This specification describes instead of writing all configuration details into
+file, Store everything in etcd and configure all tendrl modules via api request.
+
+== Problem description
+
+Currently, All the tendrl modules are configured from same file. Also, while
```
mbukatov, line 10: Yeah, all the tendrl modules are configured via `tendrl.conf`, but that was your design choice.

mbukatov:
What would be left in `tendrl.conf` then?

cybernth (commented):
As per the comments https://github.com/Tendrl/alerting/pull/1#issuecomment-264384626  and https://github.com/Tendrl/specifications/pull/11#issuecomment-264385532 @brainfunked , this is my understanding:
* Common  module ships tendrl.conf. Any components like access etcd must utilize this config file via common module
* All other configurations including logging will be centralised and stored on etcd. On startup, each component will push the default configuration to etcd
* Any further configuration changes must be done through APIs exposed 
* Restart of components should not overwrite the configration on etcd

@brainfunked  comments?

brainfunked:
@nthomas-redhat I've read through the discussion here. I can elaborate in detail. But before that, would you mind listing down what specifically is being configured via tendrl.conf right now, so that I can provide specific examples in my explanation?

cybernth:
@brainfunked , currently tendrl.conf has common, node_agent  and sds specific configrations. I have added the snippet from the configuration file here for reference:

[common]

etcd_port = 2379

etcd_connection = 10.70.42.121

log_cfg_path = /etc/tendrl/common_logging.yaml
log_level = DEBUG


[node_agent]
log_level = DEBUG
tendrl_exe_file_prefix = /tmp/.tendrl_runner
log_cfg_path = /etc/tendrl/node_agent_logging.yaml

deployment_type = production

[gluster_integration]
log_level = DEBUG
log_cfg_path = /etc/tendrl/gluster_integration_logging.yaml

[ceph_integration]
log_level = DEBUG
log_cfg_path = /etc/tendrl/ceph_integration_logging.yaml
crush_host_type = host
crush_osd_type = osd
favorite_timeout_factor = 3
server_timeout_factor = 3
cluster_contact_threshold = 60
cluster_map_retention = 3600

brainfunked:
First and foremost, an application has layers of configuration. These layers determine which configuration options are overriden. For a generic CLI application, there are two layers: configuration file and cli options/environment variables. Configuration files are used for long term configuration of the application while the cli options or environment variables allow temporary override of some or all of the configuration parameters at the time of execution.

Tendrl being a distributed deployment, composed of different components. We could have the following hierarchy of the configuration layers:
1. Global defaults in deployment-wide configuration in the central store
2. Cluster-wide configuration, which overrides the global defaults for any specific cluster, in the central store
3. Command line parameters and environment variables get stored in etcd, per component. These are intended to be used for purposes where some information needs to be provided at the deployment time, per component, that cannot be obtained any other way.
4. Temporary component specific overrides, as an administrator, are easier when done via configuration files on the individual nodes. For instance, making one of the node agents log debug information, while the cluster-wide configuration is to log at info. We could revisit this in the future, but for now my preference would be to allow temporary overrides from the configuration files, which are not reflected in the central store.

However, given the limited number of configuration options currently needed, I think we could implement the following strategy as part of this specification:

## Deployment-wide configuration

Deployment-wide configuration needs to be in the central store, configured via the API. UI can use the API calls to make these configuration changes. The configuration in the central store needs to be treated as an object hierarchy, which appropriate atoms and flows to allow modification of this object hierarchy. This type of configuration could include:
* Logging configuration (syslog configuration, priority, filter level etc.). Also reference https://github.com/Tendrl/specifications/issues/28
* Installation source for components tendrl would provision (pip, rpm etc.)
* Runtime system directory paths
* Paths to any local data stores required by components

## Commons configuration

The commons library provides the connection modules for any component to connect to and work with the central store. This library needs a configuration file to be stored on disk. This connection information should end up in `/etc/tendrl/tendrl.conf`.

## Integration configuration

Integration specific configuration should be under cluster specific hierarchy in the central store.

## Component specific temporary administrative overrides

Each component could have it's own configuration file under `/etc/tendrl/` that could allow specific overrides, for instance to logging priority etc. The exact parameters that could be overridden this way need to be determined on a case-by-case basis for every component. The only requirement is that they be documented as part of any pull request that proposes to implement them.

cybernth:
@brainfunked LGTM
@GowthamShanmugam please make appropriate modifications to the spec file based on this

nnDarshan:
@GowthamShanmugam Please try to address https://github.com/Tendrl/node_agent/issues/99 as part of this

GowthamShanmugam:
@nthomas-redhat @shtripat @nnDarshan @brainfunked please review it

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,246 @@
...
+
+Centralized deployment-wide or cluster-wide configuration in the central store,
+which can be overridden per component via configuration files or command line
+options.
+
+== Proposed change
+
+* Modify the sample configuration files in each modules. It should contain
+  its own and relevant configuration information only.
+* Rename the sample configuration files in each modules with proper name except
```
shtripat: Common library should not have conf sample and the same would be taken care as part of re-factoring task. The module which uses common module, would impose its configurations to common module while logging etc.

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,246 @@
...
+== Data model impact
+
+* Add config details in etcd with key config/repo_name.
+* Add integration specific details in etcd with key config/cluster_id/repo_name.
+
+== Impacted modules
+
+=== Tendrl API impact:
+
+* Api to get deployment wise configuration details v2/keys/config/repo_name.
```
shtripat: Add the output structure for these apis as well.

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,246 @@
...
+== Other deployer impact
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
```
shtripat, line 185: Here you should add only the specs issue URL and rest should be added in "References" section

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: i will modify  it

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: i will add it

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: done

GowthamShanmugam:
As per @shtripat comment only node_agent have its own configuration details as well as etcd configuration details also

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,232 @@
...
+* Modify tendrl definition file in node_agent to store node_agent specific
+  configuration details.
+
+== Alternatives
+
+None
+
+== Data model impact
+
+* Add config details in etcd with key config/repo_name.
```
r0h4n: This should go in "/_tendrl/config/repo_name" and not config/repo_name

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,232 @@
...
+* Add integration specific details in etcd with key config/cluster_id/repo_name.
```
r0h4n: This should go in /clusters/cluster_id/_tendrl/config/repo_name and not config/cluster_id/repo_name

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,232 @@
...
+* Create atoms and flows to allow modification on configuration in etcd. It
+  should modify local configuration file also.
+* Modify tendrl definition file in node_agent to store node_agent specific
+  configuration details.
+
+== Alternatives
+
+None
+
+== Data model impact
```
r0h4n, line 73: Define and implement  a "Config" object for each Tendrl component

r0h4n (changes requested):

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: i will change it

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: i will change it

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: i will change it

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+* Define and implement a "Config" object for each Tendrl component
+* Add config details in etcd with key /_tendrl/config/repo_name.
```
r0h4n: "/_tendrl/Config/repo_name"

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,234 @@
...
+* Add integration specific details in etcd with key /clusters/cluster_id/
+  _tendrl/config/repo_name.
```
r0h4n: "/clusters/cluster_id/_tendrl/Config/repo_name"

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+== Impacted modules
+
+=== Tendrl API impact:
+
+* Api to get deployment wise configuration details v2/keys/config/repo_name.
```
r0h4n: update v2/keys/config/repo_name to "/_tendrl/Config/repo_name"

r0h4n (changes requested):

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: done

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: done

specs/separate_configuration_files_per_component.adoc
GowthamShanmugam: done

specs/separate_configuration_files_per_component.adoc
r0h4n: Refer https://github.com/Tendrl/specifications/pull/20/files#diff-a7f219a4f7014686f05ffba6d50dbdcdR250

We need similar Config object

r0h4n (changes requested):

specs/separate_configuration_files_per_component.adoc
```diff
@@ -0,0 +1,234 @@
...
+file should not affect configurations in etcd. Any changes in integration
+specific configuration in etcd will affect config files of particular cluster
+nodes only. But changes in other configuration in etc will affect all nodes
+local config files.
+
+== Use Cases
+
+Centralized deployment-wide or cluster-wide configuration in the central store,
+which can be overridden per component via configuration files or command line
+options.
```
mkudlej, line 45: From what I've read:
1) common configuration options will be stored in Etcd
2) every node can override configuration by local configuration options in file
3) there will be Rest API functions for CRUD for options in Etcd

Do I understand it right?

I have some questions:
1) What will store common options into Etcd during installation?
2) What will happen during potential update of options in Etcd and local files?
3) Will API be able to change options stored in local files?
4) What will be preferred way to configure Tendrl components on nodes in Ceph/Gluster cluster? For example I would like to increase logging level for group of nodes.

sankarshanmukhopadhyay: @r0h4n @brainfunked could either of you respond to Martin's queries?

r0h4n: About "there will be Rest API functions for CRUD for options in Etcd"
This has not been implemented yet, and IMO this should be on a very low priority (in terms of our timelines) since configuration management is already solved by puppet/ansible etc per node 

Questions
1) What will store common options into Etcd during installation?
Ans: Nobody, config is required and pushed to etcd by each tendrl service on start

Once we have the API for conifgs:
2) What will happen during potential update of options in Etcd and local files?
Ans: Etcd config will be the only source of config, however during update the node local config changes will be overwritten on the etcd config

3) Will API be able to change options stored in local files?
Ans: Yes

4)What will be preferred way to configure Tendrl components on nodes in Ceph/Gluster cluster? For example I would like to increase logging level for group of nodes
Ans: Before tendrl services have started, use node local config files
Once tendrl services have started and updated config in etcd, use tendrl API to configure tendrl nodes

GowthamShanmugam: Is this still need, i think, this is already done

r0h4n:
Done:
- separated out per-component configuration files.

Yet to be done:
- tendrl-api for CRUD actions on config items

@brainfunked when should we take this up?

r0h4n:
Parking this one since no activity

PR #18, opened 2016-12-02
Pluggability of different supported alert notifying means

anmolbabu:
tendrl-bug-id: Tendrl/specifications#40


anmolbabu:
@brainfunked @r0h4n @nthomas-redhat @shtripat Please review

anmolbabu:
This has been raised in accordance with comment on https://github.com/Tendrl/alerting/pull/1

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,165 @@
...
+  alerts via tendrl-api. This api will require the details like what kinds of
+  alerts the user desires to be notified of, the medium through which user
+  desires to be notified.The appropriate validations will be made as part of
+  execution of the api job to find if all required configurations are present
+  for the choices made in api request body.
+
+----
+  ex
+      {
+        Name: User1,
```
shtripat: %s/Name/name/g

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+=== Alternatives
+
+To be explored.
+
+=== Data model impact
+
+The user management capability needs to be built into tendrl.
+Till then these configurations can be maintained as part of the config file as
+below:
```
shtripat: Do you mean to add additional users as well to the same configuration file? How do secure the password for the users in the flat files?

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,165 @@
...
+None
+
+=== Other deployer impact
+
+None
+
+
+=== Developer impact
+
+== Implementation
```
shtripat, line 381: I feel implementation details should be added for this talking about how the user details are maintained and how security etc is taken care for the details.

specs/pluggable_alert_delivery.adoc
anmolbabu: No I meant until tendrl central store has the users maintained in it, we will read the user details from the conf file.

Also, for time being even for maintaining auth credentials, we can use cert file path instead of password in conf file.
I'll write a sample for the same in my update patch

specs/pluggable_alert_delivery.adoc
anmolbabu: Done

specs/pluggable_alert_delivery.adoc
anmolbabu: The reason that I did not add this is this spec is not concerned with user management. Instead this is intended to deal with only the approach that can be taken so that the user chooses all/some of the supported ways of alert delivery and also will be able to choose the types of alerts he likes to receive and accordingly how the framework can be flexible to heed to the  users choices. The maintainance of users as part of conf file is only a make-shift arrangement until the user management is built into tendrl. And this make shift arrangement will not support multiple users and will be capable of handling only 2 users the admin(alert sender) and recipient(alert receiver).Adding implementation details for the other part i.e, how handlers get involved etc...

brainfunked:
As one of the comments rightly points out, this spec is not concerned with user management, but instead with delivery mechanisms. However, there aren't details towards the latter here. What I'd like to see in this spec are details regarding how the alerting framework will handle:
* pluggable delivery modules
* expectations regarding the guarantee of the delivery of alerts and the corresponding implementation required
** do the guarantees change depending upon the importance of an alert?
** a way to assign importance to alerts of a specific type
** a possible schema for alerts such that their type can be made known
* design of the codebase to allow addition of new modules without impacting the main alerting codebase
* impact (if any) on the definition files)
* any additional flows required
* how the configuration of the alerting system will be done centrally in the central store
* how the API can discover the alerting mechanisms available.

The association of user preferences with a specific alerting module can be moved to a different spec altogether, since that spec will require a spec regarding user handling in tendrl to be accepted first.

anmolbabu:
@brainfunked Any alert in the alerting module is guaranteed to be delivered to destination regardless of degree of its critical-ity/ importance. Do we require this importance based delivery guarantee and how can that be achieved.

brainfunked:
@anmolbabu This isn't a hard requirement for me, but more of a "should we consider this?". It is possible that in certain cases, a lot of alerts are generated, not all of which are of equal importance. In general, UI would probably need to display all of the generated alerts. However, alerts via SNMP or email may not need to include each and every alert. At the same time, it could also be required to ensure that these important alerts are always guaranteed to be delivered, while it's OK to drop some of the less important, informational, alerts.

The specific case I can think of is when the alerting system goes down while trying to deliver an alert, which needs to be redelivered upon recovery.

brainfunked:
I'm creating an issue to address user handling and role based access control in tendrl, which would be important for the alerting configuration later on.

UPDATE:
The user handling related aspects of this spec can be contributed to https://github.com/Tendrl/specifications/issues/23.

anmolbabu:
@brainfunked How is the alert marked as important? Should this also be configurable per cluster/per entity? How do we ensure the guarantee of delivery across crashes of the alerting module? Is it just that an alert marked as important if failed to be notified, will be reattempted once it is started back?

brainfunked:
Alert priority has to be determined by whoever generates the alert. This is why we need well-formed alerts, with schema and appropriate metadata.

We should have a pre-defined priority scheme to which the alerts must conform. This could be similar to syslog's levels.

You'll need some kind of persistence and transactional delivery to enable guaranteed alert delivery.

brainfunked:
@anmolbabu This specification should deal only with the alerting mechanism itself, while the user handling related parts could be moved under https://github.com/Tendrl/specifications/issues/23.

specs/pluggable_alert_delivery.adoc
shtripat: ok. As discussed its better to follow cert way by default may be.

specs/pluggable_alert_delivery.adoc
shtripat: ok

shtripat:
Regarding priority of the alerts and marking them as notifications, I feel not all the alerts are notifocations and based on certain configurations or while creation of the alert we can decide if the alert is meant to be a notification and only those alerts are notified to configured recipient.

specs/pluggable_alert_delivery.adoc
```diff
@@ -35,16 +37,37 @@ via mediums such as:
...
   do is add itself as a means to list of supported notification mechanisms in
   etcd.
-* For the api to get list of the supported notification mechanisms, it can
-  look into this list.
 * The alerts raised are persisted in etcd before sending out notifications in
   '/alerts' directory of etcd by node-agent.
-* The alerts marked as severity critical are deemed important and hence their
+* Each alert will carry an additional attribute that carries its significance
+  which will be added by the respective handler. The supported significance
+  levels can be 'HIGH', 'MEDIUM' and 'LOW'.
```
cybernth: Why do we need to have separate attribute handle this. Can we use the severity itself to handle this?
critical -> notification with delivery guarantee
info -> no notifications
anthing between info and critical are notifications without delivery gurantee

specs/pluggable_alert_delivery.adoc
```diff
@@ -107,6 +132,7 @@ and for a status based alert:
   },
   'Type': 'status',
   'severity': 'Critical',
+  'significance': 'HIGH',
 }
```
cybernth, line 209: Why to have different schema for different type of events? Try to have single schema for all where the info specific to the types can go into tags

cybernth (changes requested):
couple more of things:
snip of the definitions file contains the flow should be added to the spec
Info about available alerting handlers? How to enable/disable specific alert handlers. How to configure alert handlers(For example snmp requires the ip and community details of the sender)

specs/pluggable_alert_delivery.adoc
anmolbabu: No its not 2 different schema even currently. Its just 2 different examples shown above.

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok

anmolbabu: No I feel we might require an additional field for this because consider the case below:

Lets say memory utilization on a node crossed threshold and hence CRITICAL severity alert was raised and it was notified to user. But within a while the memory utilization went below the thresholds i.e, it has become normal, now the severity of new alert is INFO while its significance is still HIGH so that user is notified of the latest state of the system and he doesn't remain with the assumption that the system's memory has crossed threshold.

anmolbabu:
@nthomas-redhat 

>couple more of things:
>snip of the definitions file contains the flow should be added to the spec

Ok

>Info about available alerting handlers? 

Ok. I'll add a snippet of definitions file to achieve this. Its just the list is in etcd and we just need an atom to fetch this list from etcd.

>How to enable/disable specific alert handlers. How to configure alert handlers(For example snmp >requires the ip and community details of the sender)

There's no action allowed on alert handler. Its just that the user who desires to be alerted by a specific means needs to have as part of his details in etcd, the configuration corresponding to the desired means of being alerted. I have not added it here because I feel its related to user handling and might need to be part of https://github.com/Tendrl/specifications/issues/23.

This spec deals with how a new handler can be hooked into without much changes in alerting framework.

r0h4n (approved):

specs/pluggable_alert_delivery.adoc
cybernth: make sense

specs/pluggable_alert_delivery.adoc
```diff
@@ -195,13 +259,15 @@ Primary assignee:
...
+     time, acked/not acked, alert type, severity, resource and significance.
+
 
 == Testing
 
-This adds new apis for user configuration which need to be tested.
+This spec introduces an api to list available means of alert notification which
+needs to be tested.
 
 == References
```
cybernth, line 410: Issues linked to this spec are not correct. Need to:
- raise an issue on specification repository
- raise issues on all the affected components and link it here

specs/pluggable_alert_delivery.adoc
```diff
@@ -195,13 +259,15 @@ Primary assignee:
 == Documentation impact
 
 * This adds 2 apis:
-  ** One to list supported alert notifying mechanisms.
-  ** One for Users to subscribe to interested alerts and be notified by chosen
-     alerting means.
+  ** Api to get list of currently supported means of notification.
```
cybernth: specify the API URL and get it reviewed by @anivargi

cybernth (changes requested):
Regarding https://github.com/Tendrl/specifications/pull/18#issuecomment-265986477
I agree that for email handler, it is linked with the user. But in case of snmp you need to configure the details of the legacy system where you want to forward the alerts to. How you are planning to handle this?

specs/pluggable_alert_delivery.adoc
shtripat: This is something we should be handling as clearing event for an existing previous critical event associated with it. Can you think of something on this line??

cybernth (approved):
Looks good to me.
Please get the APIs reviewed by @anivargi

cybernth:
@Tendrl/qe , please have a look

anmolbabu:
@anivargi please review

shtripat (commented):
Looks good to me as long as all the missing points added

r0h4n (approved):

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,371 @@
...
+
+----
+GET /alerting
+
+Sample Response:
+
+Status: 200 OK
+{
+  email,
+  snmp
```
anivargi: The response format looks off.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,371 @@
...
+}
+----
+
+  ** Api to get list of alerts with various filtering options such as based on
+     time, acked/not acked, alert type, severity, resource and significance.
+
+----
+GET /alerts/severity='CRITICAL'
```
anivargi: No need for single quotes in query parameters

anivargi (changes requested):

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok updating it to :

GET /alerting/supported_notification_medium

Sample Response:

Status: 200 OK
{
  notif_medium: [email, snmp]
}

Is the above ok?

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,370 @@
...
+
+The tendrl api needs to proxy to apis exposed by Tendrl/alerting as mentioned
+in section below(Notifications/Monitoring impact)
+
+==== Notifications/Monitoring impact:
+
+Tendrl/alerting needs to implement flows for apis as described in section
+"Tendrl API impact".
+
+The flow definition for the above will look like:
```
r0h4n, line 260: How are you handling alerts with a cluster context and not just node context?

r0h4n (changes requested):

specs/pluggable_alert_delivery.adoc
anmolbabu: For cluster context alerts, the cluster context will be added to the tags attribute which is currently wrongly mentioned as of type string. It is instead supposed to be json. And the framework is otherwise agnostic of whether its a cluster specific alert or just a node-apecific alert

r0h4n: Mention examples of how alerts look under 1. cluster context, 2. global context

r0h4n (changes requested):

r0h4n (approved):

r0h4n:
@Tendrl/tendrl-core @Tendrl/tendrl_api @Tendrl/qe please ack

cybernth (approved):

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+
+== Problem description
+
+* The different tendrl supported alerting means need to be pluggable in the
+  sense that adding any new alert notifying means should be an easy task and
+  should not involve much changes in the core alerting framework.
+* The alerting mechanism should have a way to set importance of alerts so that
+  the alerts desingated as important will be notified to users even if the
+  alerting module crashed and hence the alert could not be notified but will
+  be alerted once the alerting module is started back.
```
brainfunked, line 30: We'll also need to make the alert destinations configurable, possibly based on priority. However, that would need an entirely different specification under https://github.com/Tendrl/specifications/issues/23.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+
+== Use Cases
+
+* Dynamic addition of new means of notifying an alert.
+
+== Proposed change
+
+* The alerts raised are persisted in '/alerts' directory of etcd before sending
```
brainfunked, line 38: Do the alerts need to be scoped or can every alert be global? Would all the alerts being global enable delivery of alerts per cluster to a separate destination?

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+  out notifications. The flow is as below:
+  ** One socket is connected to for writes from collectd, read-only from the
```
brainfunked: Should probably read:

> Node agent would expose an alerts socket. It would read from this socket. The bridge and collectd would write to this socket.

Why the bridge needs to read from the socket is explained below, so no need to include that information here.

brainfunked: On a separate note, how do we handle reading from sockets? Is this done in a green thread or some such which polls the socket every now and then or is there an eventing mechanism in place? @r0h4n ?

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+     node agent and read-write from the bridge.
+  ** The generator(collectd or bridge) of the alert needs to put the alert on
+     this socket.
+  ** Any alert on that socket is always taken to etcd by the node agent
+  ** Bridge can read the alerts and act on only the ones it can act on, ignore
```
brainfunked: > In addition to writing the generated alerts, the bridge also reads from the socket. It then filters and chooses only the alerts it can act on..

and so forth..

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+     the rest it can also generate it's own alerts and put them on the socket
```
brainfunked: We've already stated above that the bridge will be generating the alerts, no need to repeat.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+     for node agent to transport to etcd
+* Alert meta-data:
+  ** Each alert will carry additional attribute that carries its significance
+     which will be added by respective handler. The supported significance
+     levels can be 'HIGH', 'MEDIUM' and 'LOW'.
```
brainfunked: The level of the alerts would be hardcoded logic in the codebase that generates the alerts.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+    *** The 'LOW' significance level alerts will not be notified. But will
```
brainfunked: This is not a valid assumption. This needs to be configurable. Some users may want to see all the alerts being generated. This cannot be hardcoded in the code. Even if it's not configurable presently, expose it in the flows in such a way that it can later be made configurable. I have a feeling that this will eventually be bubbled up into https://github.com/Tendrl/specifications/issues/29.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+        feature in the list of alerts in etcd.
+    *** The 'MEDIUM' significance level alerts will be attempted to be notified
+    *** The 'HIGH' significance level alerts will be re-attempted to be
+        notified(on restart) if a previous attempt failed for some reason.
+        (crash of alerting service).The alerts marked as significance HIGH are
+        deemed important and hence their delivery is guranteed even in case of
+        sudden crash of alerting module. This is achieved by updating a field
+        ('notified': True) of the alert in etcd directory to indicate that the
```
brainfunked: Given the pluggable nature of our transports and the possibility that each alert may need to be delivered via more than one transport.

The alerting module is updating the alerts in any case. So, I think we can expand that a bit and have a field called `delivery` or some such added by default to each alert. The alerting module would add individual endpoints there, for example:

`
delivery: {
  smtp: {
    endpoint: 'foo@bar.com',
    pickedup: True,
    delivered: False
  },
  snmp: {
     endpoint: 'foobar',
     pickedup: True,
     delivered: False
  }
}
`

This brings up another point. How do we determine if there are unsent alerts? That's where the `notified` field you've used comes in. If you notice the `pickedup` fields above, they're set to True. The reason is that I'm assuming that the alerting module picks up each alert as soon as it comes across it. At which point it adds the entire `delivery` field. At this point I'm also assuming that it immediately invokes all the applicable delivery modules to deliver this alert. `pickedup` indicate that this was done. `delivered` is updated once the delivery module confirms an alert to have been successfully delivered. Each time the alerting module goes to update the value of the `delivered` key for a specific delivery module, it can update the global value if all the nested `delivered` as True.

There are additional concerns I can think of at this point. Are the delivery modules just point in time invocations, so that they have no state of their own? In which case, how do you determine the last attempted delivery time? This goes towards the question of figuring out which alerts need to be redelivered. Of course, I think this wouldn't be a problem if every delivery is a blocking, atomic call. It either succeeds or fails. In any case, you might want to put some interval before attempting redelivery and track the number of delivery attempts made as well.

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+        alert has been notified. When alerting module is started, it scans all
+        alerts in '/alerts' directory and sends out notifications for those
+        that are marked significance 'HIGH' but are not set with
+        'notified':True
+  ** An alert will also have a field to ack it. Acking an alert can be done by
```
brainfunked: :+1:

specs/pluggable_alert_delivery.adoc
```diff
@@ -0,0 +1,372 @@
...
+        medium name. ex: email_handler for mail notification in the directory
+        alerting/tendrl/alerting/notification/
+    *** Whenever a new handler is added, the first thing the handler is
+        supposed to do is add itself as a means to list of supported
+        notification mechanisms in etcd in '/notification_medium' directory.
+    *** The alerting module will expose apis for managing configurations and
+        subscriptions for each of the notification medium which will be proxied
+        to by tendrl-api.
+        The way the api works is as follows:
+        **** The configuration if applicable will link to the tendrl maintained
```
brainfunked: I think we can implement this alerting specific way of configuration for now, but eventually, I think it should be tackled via https://github.com/Tendrl/specifications/issues/29.

brainfunked (changes requested):

specs/pluggable_alert_delivery.adoc
anmolbabu: Currently it is implented as a greenlet server that continuously polls socket

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok

specs/pluggable_alert_delivery.adoc
anmolbabu: My plan was to have per cluster and global kind of deliveries option probably it got removed while removing the user specifics from this spec. I'll add a note on some of the configs that applies to any kind of notification like the one that you just pointed.

specs/pluggable_alert_delivery.adoc
anmolbabu: Done

specs/pluggable_alert_delivery.adoc
anmolbabu: Now that the config are separated out from the user details, I'll add a note in this spec

specs/pluggable_alert_delivery.adoc
anmolbabu: ok

specs/pluggable_alert_delivery.adoc
anmolbabu: ok

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok I'll make this explicit

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok

specs/pluggable_alert_delivery.adoc
r0h4n: I have asked @anmolbabu (in previous reviews) to use Co-op Socket (http://www.gevent.org/gevent.socket.html) which is part of our consumption of green threads (via gevent library in python)

Please dont use normal processes/threads/sockets for any Tendrl functionality written in python,  Use the gevent library (http://www.gevent.org/)

specs/pluggable_alert_delivery.adoc
anmolbabu: Ok

anivargi (approved):

brainfunked (approved):

PR #19, opened 2016-12-06
Specification: Adding network details in hardware inventory

GowthamShanmugam:

### DIFF
--- a/specs/network_details_for_hardware_inventory.adoc
+++ b/specs/network_details_for_hardware_inventory.adoc
@@ -1,0 +1,183 @@
+= Specification for adding network details in node hardware inventory
+
+Add network details also as a part of node hardware inventory
+
+== Problem description
+
+Hardware inventory details are used to show exact status of the particular
+node. Currently network details are missing in node hardware inventory.
+we have to add network details like network_id, ipv4, ipv6, netmask, status,
+interface_id, sysfs_id, device_link, interface_type, model, driver_modules, drive
+hw_address, link_detected as a part of node hardware inventory.
+
+== Use Cases
+
+Tendrl should show the current status of network interface.
+
+== Proposed change
+* From node agent collect all network interface details in node inventory list and
+  store it in etcd. (Detailed information in implementation section).
+
+== Alternatives
+
+* We can also use ethtool for getting network details. But ethool is not
+  available for pip.
+
+== Data model impact
+
+* Add network details in etcd with key nodes/node_id/Networks/{interface_name}. (node level)
+* Add network details in etcd with key networks/{subnet}. (global level)
+
+== Impacted modules
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+* Add a new function in pull_hardware_inventory to collect network interface
+  details.
+* Create a new class called Network in persistence module to keep the network
+  details.
+* Create a new function in Network class to convert object into json. (We are string
+  json by json in etcd).
+* Modify the manager modules to read and store network interface details.(stroring
+  network details in node level and global level also).
+* Create an object in tendrl_definitions_node_agent for network details.
+* Commands used to retrieve network details are:
+```
+    interfaces = netifaces.interfaces()
+    ipv4_addr_list = netifaces.ifaddresses(interfaces[integer_value])[AF_INET]
+    netmask = ipv4_addr_detail[integer_value]['netmask']
+    ipv6_addr_list = netifaces.ifaddresses(interfaces[integer_value])[AF_INET6]
+```
+* Commnad for check network interface is up or not
+```
+    Command({"_raw_params": "cat /sys/class/net/wlp3s0/operstate"})
+    Command({"_raw_params": 'hwinfo --network'})
+```
+* Code to find subnet using ip address and netmask
+```
+import netaddr
+cidr = netaddr.IPNetwork('%s/%s' % (address, netmask))
+network = cidr.network
+```
+* Expected format for the network details in etcd as node level is:
+```
+/nodes/{node_id}/Networks
+
+{"interface_name": {"ipv4": ["ipv4address", ...],
+                    "ipv6": ["ipv6address, ..."],
+                    "netmask": ["subnet", ...],
+                    "subnet": "subnet",
+                    "status":"up/down",
+                    "interface_id": "",
+                    "sysfs_id": "",
+                    "device_link": "",
+                    "interface_type": "",
+                    "model": "",
+                    "driver_modules": "",
+                    "drive": "",
+                    "hw_address": "",
+                    "link_detected": ""
+                    }, ...
+}
+```
+*  Expected format for the network details in etcd as global level is:
+```
+/networks
+
+{"subnet1": [{"interface_name": "interface",
+            "ipv4": ["ipv4address", ...],
+            "ipv6": ["ipv6address, ..."],
+            "netmask": ["subnet", ...],
+            "status":"up/down",
+            }, ...
+          ],
+ "subnet2": [{"interface_name": "interface",
+             "ipv4": ["ipv4address", ...],
+             "ipv6": ["ipv6address, ..."],
+             "netmask": ["subnet", ...],
+             "status":"up/down",
+             }, ...
+           ],
+}
+```
+* Add new object in tendrl defenition file with list.
+```
+list: nodes/$Node_context.node.id/Network
+list: networks/$subnet
+```
+
+=== Tendrl/gluster_integration impact
+
+None
+
+=== Tendrl/ceph_integration impact
+
+None
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+We might have to monitor the status of the network interface(up/down).
+
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
+* Netifaces and hwinfo packages are used to find the network interface details.
+* Add netifaces and hwinfo packages as dependency.
+* From node_agent add a new function in pull_hardware_inventory to find network
+  interfaces details except 'lo' and update those details into node_inventory list.
+* Create a new class called "Network" in persistence module to keep network details.
+* Add a new function in Network to convert object to json.
+* Modify the manager to populate the instance of newly created class (network) using
+  node_inventory list and persist the instance in etcd.
+* Create an object in tendrl_definitions_node_agent for network details.
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* git-hub issue: https://github.com/Tendrl/node_agent/issues/78
+
+== Dependencies
+
+* Netifaces, netaddr and hwinfo are  dependency to find network details
+* Bonded network interface details will be handled in another specification file
+
+== Testing
+
+* Sanity check for flow.
+* Check all network details are persisted successfully.
+
+== Documentation impact
+
+None
+
+== References
+
+* Bonded interface details will be handled in another specification file


GowthamShanmugam:
@shtripat @r0h4n @shtripat @nnDarshan please review it

GowthamShanmugam:
@r0h4n @shtripat @nnDarshan please review

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+* Modify pull_hardware_inventory.py to find active network interfaces details
+  except lo, virbr0, virbr0-nic and update those details into node_inventory
+  list.
+* Create a new class called Network in persistence with respective attribute.
+* Modify manager.py to populate network details from node_inventory list to
+  Network object.
+* Modify persister.py to persist the Network object in etcd.
+
+== Alternatives
+
```
nnDarshan, line 22: you can mention about the few other network-utilities that you referred, also provide info about why you choose this over that.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+* Netifaces package is used to find active network interface in particular
+  node.
+* Add netifaces in requirements.txt and tendrl-node-agent.spec as a dependency.
+* Modify pull_hardware_inventory.py to find active network interfaces details
+  except lo, virbr0, virbr0-nic and update those details into node_inventory
+  list.
+* Create a new class called Network in persistence with respective attribute.
+* Modify manager.py to populate network details from node_inventory list to
+  Network object.
+* Modify persister.py to persist the Network object in etcd.
```
nnDarshan: Its a bit implementation specific, can you have more like an overview of what you are going to change ?

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+None
+
+=== Tendrl/node_agent impact
+
+* tendrl/node_agent/manager/manager.py
+* tendrl/node_agent/manager/pull_hardware_inventory.py
+* tendrl/node_agent/persistence/persister.py
+* requirements.txt
+* tendrl-node-agent.spec
+
```
nnDarshan, line 112: can you write a couple of sentences about the impact, instead of mentioning the file names ?

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+=== Tendrl/ceph_integration impact
+
+None
+
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
```
nnDarshan, line 125: we might have to monitor the status of the network interface, and alert when interfaces goes down of comes up. May be in future when alerting framework is ready.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+None
+
+== Implementation
+
+** github issue link here
+
+== Assignee(s)
+
+Primary assignee: TODO
```
nnDarshan: mention your git hub name here, as you have already sent a PR for this

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+* Modify manager.py to populate network details from node_inventory list to
+  Network object.
+* Modify persister.py to persist the Network object in etcd.
+
+== Alternatives
+
+None
+
+== Data model impact
+
```
nnDarshan, line 27: you can mention about the path where this details is stored in etcd ?

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,115 @@
...
+* Adding network details in node hardware inventory
+
+== Impacted modules
```
nnDarshan, line 31: I think you have missed api impact section, there you can mention that api for listing network interface to be added by api-developer.

nnDarshan (changes requested):

GowthamShanmugam:
@nnDarshan @shtripat @r0h4n @Tendrl/qe please review it

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,112 @@
+= Specification for adding network details in node hardware inventory
+
+Add network details also as a part of node hardware inventory
+
+== Problem description
+
+Hardware inventory details are used to show exact status of the particular
+node. Currently network details are missing in node hardware inventory.
+we have to add network details like IPv4, IPv6, netmask as a part of node
```
cybernth: Are we dealing with ipv6 now?

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+== Data model impact
+
+Add network details in etcd with key nodes/node_id/Network/interface_name
+
+== Impacted modules
+
+=== Tendrl API impact:
+
+Api to get network details v2/keys/nodes/node_id/Network
```
cybernth: Do we really need to have an API to get the network of the node?. I don't think so. Rather this should be part of Node and retrieved through GetNodeList api

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+Modify persistence and manager modules to find and store active network
+interface details.
+
```
cybernth, line 112: What about the definitions? Do we have an object defined for network? 
Also any new atoms needs to be added?

cybernth (changes requested):

GowthamShanmugam:
@nthomas-redhat ipv6 is not required but i feel it is better to keep those details for future use

GowthamShanmugam:
@nthomas-redhat @nnDarshan please review it

cybernth:
@GowthamShanmugam , I am not seeing the changes I have requested. Also looks like the latest version posted here is not right?

specs/network_details_for_hardware_inventory.adoc
shtripat: Thats correct. We would need an object defined for sure.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,131 @@
...
+  network interfaces details except lo, virbr0, virbr0-nic and update those
+  details into node_inventory list.
+* Create a new class in persistence module to keep network details.
+* Add a new function in persister to persist network details.
+* Modify the manager to populate the instance of newly created class using
+  node_inventory list and persist the instance in etcd.
+
+== Alternatives
+
+* Netifaces package is used to find active network interface in particular
```
shtripat: This is talking about what mechanism you have used. Rather @nnDarshan meant you to add various tools which were evaluated and could be used for getting the details I feel.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,131 @@
...
+  node.
+* Add netifaces package as dependency.
+
+== Data model impact
+
+Add network details in etcd with key nodes/node_id/Network/interface_name
```
shtripat: Also definitions should get an object called Network I feel

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,131 @@
...
+
+None
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+We might have to monitor the status of the network interface(up/down).
```
shtripat, line 127: @anmolbabu do we need more detail to be captured here??

GowthamShanmugam:
@nthomas-redhat i will remove IPV6 details from this, and i will add about modification in tendrl_definition

GowthamShanmugam:
@nthomas-redhat atoms are not required for this

GowthamShanmugam:
@nthomas-redhat @shtripat @nnDarshan all changes are done.. please review it

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,135 @@
...
+we have to add network details like IPv4, netmask, status as a part of
+node hardware inventory.
+
+== Use Cases
+
+Tendrl should show the current status of active network interface.
+
+== Proposed change
+
+* From node_agent add a new function in pull_hardware_inventory to find active
```
nnDarshan: Mention about the tool that you will be using to get network details

GowthamShanmugam:
@nnDarshan done

nnDarshan (commented):
Looks Good to me. @anmolbabu Could you please comment on the monitoring impact

specs/network_details_for_hardware_inventory.adoc
anmolbabu: That's correct. Whenever a state change is detected, the same needs to go to etcd's /alerts directory and correspondingly a custom handler needs to be written to handle the event.

Ideally, the procedure is anyone who detects an alert writes it to a socket that is opened by node-agent and the node-agent writes this alert to etcd after the corresponding handler has massaged it. 

But since this alert is detected in node-agent itself, the node-agent can directly invoke an event handling gateway function(which will anyway be invoked for the messages in socket in usual path) which will take care of the rest of the things. Offcourse a custom handler to handle this event in a custom manner if required also needs to be written.

I would suggest to refer section 'Data Model Impact' of https://github.com/Tendrl/specifications/pull/18/files for the details about alert structure(Its still in review)

specs/network_details_for_hardware_inventory.adoc
anmolbabu: I think things like significance of this alert I assume it would be 'HIGH' can be captured + a sample alert can be populated here by referring to section 'Data Model Impact' of https://github.com/Tendrl/specifications/pull/18/files for the details about alert structure(Its still in review). Apart from that, from the alerting perspective I don't see any more requirement. 

But, I have 1 question here, does this need to be graphed (binary status graph with up, down values) for user to see how the interface has been behaving. If yes, we need to maintain this as well in graphite and in such a case we need to see how node-agent can push a statistic to time-series db. Probably the performance-monitoring application needs to expose apis to push stats as well.

An alternative in my opinion is that if this can be implemented as a collectd plugin emitting non-zero higher value for down status and a zero for up(only to benefit from collectd's thresholding plugin that can emit an alert for values greater than configured threshold in our case say 0 or some value greater than 0) we can benefit from the normal statistics data and alert flows. Anyway since the status to int mapping is very much hard-coded within applications in tendrl space, the corresponding interpretations can be done wherever required. 

Just that I am somehow not happy with performance-monitoring exposing a way to push any stat to graphite.

@shtripat what do you think?

shtripat: I would certainly prefer the option-2 of having a collectd plugin to push data to graphite (if we need, for sure I prefer this to be there in graphite. The only thing we need to see if real estate available for the same in dashboard)

GowthamShanmugam:
@shtripat @nthomas-redhat @r0h4n @brainfunked review please

specs/network_details_for_hardware_inventory.adoc
cybernth: @anmolbabu @nnDarshan does this needs to covered as part of this? I think it come under the scope of another spec which talks about the alerting/monitoring of disks. Does this make sense?

specs/network_details_for_hardware_inventory.adoc
cybernth: @anmolbabu @shtripat , As I commented earlier, this should be handled as part of monitoring, need not come under scope of this specification

anmolbabu: @nthomas-redhat But I feel if we are doing it from collectd then there is no need to do it from node-agent.

I feel,
* Node-agent can do an initial 1 time sync
* Thereafter, the alerting framework's disk alerts specific handler can set suitable states in etcd as well apart from the normal notification and other stuff.

r0h4n: Node agent owns these objects under its namespace, and expects them as defined under etcd.

If any other Tendrl component or external module (collectd) wants to own the object, it has to implement it as per Tendrl Definitions, meaning: it should look the same in etcd as it was defined under the node-agent namepsace.

So, keep the objects under node-agent and let node-agent manage its state in etcd.

specs/network_details_for_hardware_inventory.adoc
anmolbabu: Done

specs/network_details_for_hardware_inventory.adoc
anmolbabu: Done

anmolbabu (approved):

r0h4n (approved):

r0h4n:
@Tendrl/tendrl-core please review

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* Add netifaces as a dependency.
```
r0h4n: Is this available in epel, rhel6/7, pip?

r0h4n (changes requested):

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: yes available, for rhel, centos_epel: python-netifaces,  pip: netifaces

r0h4n: Ok thanks

r0h4n (approved):

r0h4n:
@Tendrl/tendrl-core please approve

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,132 @@
...
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
```
r0h4n, line 138: There's a new dependency being added, please document impact @nthomas-redhat @TimothyAsir

r0h4n (changes requested):

specs/network_details_for_hardware_inventory.adoc
cybernth: @GowthamShanmugam , Can you update this section with packaging related changes required. Pm me if you need any help here

GowthamShanmugam: ok i will update

GowthamShanmugam: i will add it Dependency section

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,133 @@
...
+node. Currently network details are missing in node hardware inventory.
+we have to add network details like IPv4, netmask, status as a part of
+node hardware inventory.
+
+== Use Cases
+
+Tendrl should show the current status of network interface.
+
+== Proposed change
+* Netifaces is used to find the network interface details.
```
brainfunked: What is `Netifaces`? Is it a class in the code, a type of object to be used in the flows, some command?

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,133 @@
...
+* Add netifaces as dependency.
+* From node_agent add a new function in pull_hardware_inventory to find network
+  interfaces details except lo, virbr0, virbr0-nic and update those details
```
brainfunked: You may exclude the loopback interfaces, but the exclusion of the other interfaces must not be hardcoded. If necessary this should be a global configuration option somewhere.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,133 @@
...
+* Create a new class in persistence module to keep network details.
+* Add a new function in persister to persist network details.
+* Modify the manager to populate the instance of newly created class using
+  node_inventory list and persist the instance in etcd.
+* Create an object in tendrl_definitions_node_agent for network details.
+
+== Alternatives
+
+* We can also use ethtool for getting network details. But ethool is not
+  available for pip.
```
brainfunked, line 24: `ethtool` is an external command. I don't see how it needs to be "available" for `pip`.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,133 @@
...
+* Netifaces is used to find the network interface details.
+* Add netifaces as dependency.
+* From node_agent add a new function in pull_hardware_inventory to find network
+  interfaces details except lo, virbr0, virbr0-nic and update those details
+  into node_inventory list.
+* Create a new class in persistence module to keep network details.
+* Add a new function in persister to persist network details.
+* Modify the manager to populate the instance of newly created class using
+  node_inventory list and persist the instance in etcd.
+* Create an object in tendrl_definitions_node_agent for network details.
```
brainfunked: What's the plan for refreshing the network inventory after the initial import?

specs/network_details_for_hardware_inventory.adoc
brainfunked: Alerting is just a part of the story. What about refreshing our inventory when the hardware changes, in general? Alerting is a side effect of these changes. The first impact must always be on our state store itself. Theoretically it's possible to set monitors on the changes in the state store and generate alerts there, but that's a different issue altogether. I'll review the alerting specification and see if it answers my question.

specs/network_details_for_hardware_inventory.adoc
brainfunked: @anmolbabu Speaking of monitoring, we should track the network utilisation data per interface. It'll be useful for tracking down bottlenecks later.

brainfunked (changes requested):

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: netifaces is a package

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: Every 3 min once we are getting node inventory details and compare it with local inventory file, If there is any change then it will updated into etcd. So network details also updated.

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: ethtool is an package, it is also giving network details but it is not available for pip

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: i will add it virbr0 and virbr0-nic

specs/network_details_for_hardware_inventory.adoc
brainfunked: I reviewed the alerting spec. It does not answer my question. How are we tracking changes to the hardware inventory? We may need a short and a long term solution to this problem. @anmolbabu @nthomas-redhat

GowthamShanmugam:
@brainfunked done all the changes, please review

specs/network_details_for_hardware_inventory.adoc
anmolbabu: Ok. I already have a plugin to calculate network utilization in one of my PRs. And it currently calculates both interface level and average across interfaces and puts both to time-series db.

specs/network_details_for_hardware_inventory.adoc
cybernth: @brainfunked Two ways you can handle this:
1. Process the external events send by the agents(Console-2 was depending on the events from storaged)
2. During the inventory sync, infer the change and emit alerts based on that

Utilisation data(any;network, disk etc.) collection and analysis(and generating alerts) is the responsibility of the monitoring.

shtripat: In agreement with @nthomas-redhat

julienlim (changes requested):
@GowthamShanmugam @brainfunked  Does this take into consideration bonded interfaces?  Do we collect the speed on the interface beyond IP address, netmask and state/status?

GowthamShanmugam:
@julienlim these are comes under monitoring right?

specs/network_details_for_hardware_inventory.adoc
brainfunked: :+1:

specs/network_details_for_hardware_inventory.adoc
brainfunked: I think this interval will need to be revisited over time. In general, I would prefer that we not use polling for this, but for now this could work.

GowthamShanmugam: @brainfunked But this interval is not part of this patch..it is already implemented..  In this patch i am just collecting network details and updated it

brainfunked:
@GowthamShanmugam no, bonded interfaces should be detected as part of the network inventory. If necessary, you need to invoke specific commands to detect bonded and bridged interfaces.

GowthamShanmugam:
@brainfunked netifaces will find all the network interface, including boding interface also. what are all the interfaces you can see by ifconfig or ip addr command is handled by netifaces

GowthamShanmugam:
@brainfunked I feel bonded interface handle in another specification.

GowthamShanmugam:
@nthomas-redhat @shtripat @brainfunked please review this

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,138 @@
+= Specification for adding network details in node hardware inventory
+
+Add network details also as a part of node hardware inventory
+
+== Problem description
+
+Hardware inventory details are used to show exact status of the particular
+node. Currently network details are missing in node hardware inventory.
+we have to add network details like IPv4, netmask, status as a part of
```
mkudlej: I expect that Netifaces library supports it, but I would like to ask if you count also with IPv6 support?

GowthamShanmugam: @mkudlej but we are not using ipv6 now @nthomas-redhat your suggestion please

mkudlej: @GowthamShanmugam Yes, I know but I think that upstream Ceph and Gluster support it, right?

GowthamShanmugam: @nthomas-redhat your suggestion please

GowthamShanmugam:
@nthomas-redhat @brainfunked @shtripat please review

GowthamShanmugam:
@nthomas-redhat @brainfunked @r0h4n please review

specs/network_details_for_hardware_inventory.adoc
shtripat: I feel for upstream we should collect and maintain IP6 details as well.

shtripat (commented):
LGTM as as long as ipv6 related things are finalized...

GowthamShanmugam:
yes @brainfunked told me to collect all the possible details so i added ipv6 also and some other extra details also.

r0h4n (approved):

GowthamShanmugam:
@nthomas-redhat @brainfunked  please review this

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+Tendrl should show the current status of network interface.
+
+== Proposed change
+* From node agent collect all network interface details in node inventory list and
+  store it in etcd. (Detailed information in implementation section).
+
+== Alternatives
+
+* We can also use ethtool for getting network details. But ethool is not
```
dahorak, line 23: Please be aware, that ethtool provide very limited information if launched on virtual machine for `virtio` type of NIC. I even think, that some information are not available at all for `virtio` type of NIC (for example link `Speed`).

I know, that this didn't looks like big problem for production environment (where probably all storage nodes will be physical machines), but it might be problem for testing, examination and demo purposes.

cybernth (approved):

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: @dahorak ya i accept your point, it gives very limited details only, in initial stage we excepted ipv4 , ipv6 and subnet mask details only, at that time we just tried this tools.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,150 @@
...
+                    "driver_modules": "",
+                    "drive": "",
+                    "hw_address": "",
+                    "link_detected": ""
+                   }, ...
+}
+```
+* Add new object in tendrl defenition file with list.
+```
+list: nodes/$Node_context.node.id/Network
```
brainfunked, line 109: What is the exact structure of the data in etcd? I think that there needs to be a global `networks` hierarchy as well, where we maintain a list of networks, under which we point to each node that belongs to this network. This is in addition to the network details maintained under each node.

GowthamShanmugam: @brainfunked 
Inside nodes/{node_id}/Network/{interface_name}
value : {"ipv4": ["ipv4address", ...],
                    "ipv6": ["ipv6address, ..."],
                    "netmask": ["subnet", ...],
                    "status":"up/down",
                    "interface_id": "",
                    "sysfs_id": "",
                    "device_link": "",
                    "interface_type": "",
                    "model": "",
                    "driver_modules": "",
                    "drive": "",
                    "hw_address": "",
                    "link_detected": ""
                   }
this is the structure for network details, This network details is different for the particular node. we cant give node belongs to this network. This network ipv4, ipv6 address will differ for another nodes.

brainfunked: I don't think the IP being under the interface name will scale. You'd want to allow clients to directly filter nodes based on IPs and hostnames. You'd also need to enable an index of hostnames to node ids and hostnames to IP addresses. These need to be maintained based on changes in the IP addresses on individual nodes. Essentially, I'm looking at being able to obtain network details without relying on scanning all of the nodes.

Also, my comment is specifically towards https://redhat.invisionapp.com/share/2K8M4PQYZ#/screens/188984130. Here, we need to list the networks detected, and not the IP addresses per node. This is essentially a calculation based on the netmask for each of the detected IPs. It could be done on the fly by the API, or be stored permanently in etcd, should this information need to be known elsewhere.

GowthamShanmugam: @brainfunked 
1. global structure is:

suggestion :1
 /networks/{subnet1}/{node_id1}
       Value:  [
                    {"node_id": "node_id",
                     "ipv4": ["ipv4address", ...],
                     "ipv6": ["ipv6address, ..."],
                     "netmask": ["subnet", ...],
                     "status":"up/down",
                     "interface_id": "",
                    "sysfs_id": "",
                    "device_link": "",
                    "interface_type": "",
                   "model": "",
                   "driver_modules": "",
                  "drive": "",
                  "hw_address": "",
                  "link_detected": ""
              },
                .....]

/networks/{subnet1}/{node_id2}
       Value:  [
                    {"node_id" : "node_id",
                     "ipv4": ["ipv4address", ...],
                     "ipv6": ["ipv6address, ..."],
                     "netmask": ["subnet", ...],
                     "status":"up/down",
                     "interface_id": "",
                    "sysfs_id": "",
                    "device_link": "",
                    "interface_type": "",
                   "model": "",
                   "driver_modules": "",
                  "drive": "",
                  "hw_address": "",
                  "link_detected": ""
              },
                .....]

Suggestion : 2
      (read from etcd using network like /networks/{subnet1} and  then append json in list that in list. Then dump in etcd)

 /networks/{subnet1}
                value: [
                     {"node_id" : "node_id",
                     "ipv4": ["ipv4address", ...],
                     "ipv6": ["ipv6address, ..."],
                     "netmask": ["subnet", ...],
                     "status":"up/down",
                     "interface_id": "",
                    "sysfs_id": "",
                    "device_link": "",
                    "interface_type": "",
                   "model": "",
                   "driver_modules": "",
                  "drive": "",
                  "hw_address": "",
                  "link_detected": ""
              }, ...
          ]
/networks/{subnet2}
         value: [ {"node_id" : "node_id",
                     "ipv4": ["ipv4address", ...],
                     "ipv6": ["ipv6address, ..."],
                     "netmask": ["subnet", ...],
                     "status":"up/down",
                     "interface_id": "",
                    "sysfs_id": "",
                    "device_link": "",
                    "interface_type": "",
                   "model": "",
                   "driver_modules": "",
                  "drive": "",
                  "hw_address": "",
                  "link_detected": ""
              }, ...]

In second approach we can find no.of interface like len(network1), list length is the number of interface. 
 
2. node wise structure is

/nodes/{node_id}/Network/{interface_name}
value: [ 
      {"ipv4": ["ipv4address", ...],
       "ipv6": ["ipv6address, ..."],
       "netmask": ["subnet", ...],
       "status":"up/down",
        "interface_id": "", 
       "sysfs_id": "",
       "device_link": "",
       "interface_type": "", 
       "model": "",
       "driver_modules": "",
       "drive": "",
       "hw_address": "",
       "link_detected": ""
      }
   ]

Is that global structure is ok?
whether i have to keep node wise and global wise structure or global wise structure is enaf?

GowthamShanmugam: @nnDarshan @shtripat @brainfunked

shtripat: I personally feel node level network info is required. Additionally you can have system level sub-net details

r0h4n: @GowthamShanmugam Suggestion 2 looks good, also keep both /nodes and global /networks data

GowthamShanmugam: @r0h4n ok done

GowthamShanmugam:
@r0h4n @brainfunked @shtripat @nnDarshan please review this i have updated this spec

r0h4n (approved):

specs/network_details_for_hardware_inventory.adoc
nnDarshan: I feel Suggestion 2 is better.

nnDarshan (commented):
LGTM

shtripat:
LGTM

GowthamShanmugam:
@r0h4n please merge this

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,201 @@
...
+*  Expected format for the network details in etcd as global level is:
+```
+/networks
+
+{"subnet1": [{"ipv4": ["ipv4address", ...],
+            "ipv6": ["ipv6address, ..."],
+            "netmask": ["subnet", ...],
+            "subnet": "subnet",
+            "status":"up/down",
+            "interface_id": "",
```
brainfunked: Details such as interface_id, sysfs_id etc. need not be present on a global level. On the global scope, the network information is used to identify which interfaces on which hosts belong to the network and what their IPs are, that's all. All the other details belong to specific network interfaces on specific nodes.

specs/network_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,201 @@
...
+```
+import netaddr
+cidr = netaddr.IPNetwork('%s/%s' % (address, netmask))
+network = cidr.network
+```
+* Expected format for the network details in etcd as node level is:
+```
+/nodes/{node_id}/Network
+
+{"interface_name": {"ipv4": ["ipv4address", ...],
```
brainfunked, line 70: Just make it an array of ipaddr/netmask combination. No need to store netmasks separately.

Also, store the aliases and their corresponding addresses in a hash under the aliases key or some such.

specs/network_details_for_hardware_inventory.adoc
brainfunked: I agree with @r0h4n.

brainfunked (changes requested):

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked ok

specs/network_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked ok

GowthamShanmugam:
@brainfunked all changes are done, please review

GowthamShanmugam:
@brainfunked  i have one doubt, if the interface is down then does not have ip and subnet mask, so we can't find out subnet mask, so in node wise i cant store it as subnet wise, better we store it as interface wise. I feel current implementation for node wise network details are ok.

GowthamShanmugam:
@brainfunked @r0h4n @shtripat @nthomas-redhat @nnDarshan yours suggestion please 
i feel in node wise interface wise structure is ok. In global wise we can keep subnet, because i am storing only which interfaces have subnet in global.

GowthamShanmugam:
@brainfunked @shtripat @r0h4n @nthomas-redhat @nnDarshan  If i store subnet wise in node level then i can't say interface is up/down

GowthamShanmugam:
@nthomas-redhat @r0h4n @shtripat @nnDarshan @brainfunked if we dont want status of interface (up/down) then we can store subnet wise in node level also, please give your comments

r0h4n:
@GowthamShanmugam Please figure out how to uniquely identify networks/subnets/ports(interfaces)

check https://github.com/openstack/heat/blob/master/heat/engine/resources/openstack/neutron/port.py

https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L74

or check any python module handling networks and see their data model for networks/subnets/ports

GowthamShanmugam:
@r0h4n we can merge this

PR #21, opened 2016-12-06
Specification: Adding disk details in hardware inventory

GowthamShanmugam:

### DIFF
--- a/specs/disk_details_for_hardware_inventory.adoc
+++ b/specs/disk_details_for_hardware_inventory.adoc
@@ -1,0 +1,223 @@
+= Specification for adding disk details in node hardware inventory
+
+Add disk details also as a part of node hardware inventory
+
+== Problem description
+
+Hardware inventory details are used to show exact status of the particular
+node. Currently disk details are missing in node hardware inventory. Add all possible
+disk details in node hardware inventory and store it in etcd.
+
+== Use Cases
+
+Tendrl should show the current status of disk.
+
+== Proposed change
+* From node agent collect all possible disk details in node inventory list and store
+  it in etcd. (Detailed information in implementation section).
+
+== Alternatives
+
+* hdparm is also giving some disk details like serial number and firmware but it
+  works only for ATA disks types. For SCSI disk we have to use sdparm.
+
+== Data model impact
+
+* Add used disks id in etcd with key nodes/node_id/Disks/Used.
+* Add free disks id in etcd with key nodes/node_id/Disks/Free.
+* Add all disks details in etcd with key nodes/node_id/Disks.
+* There needs to be one object per disk, which contains all the information about
+  the disk. This needs to be in the disks directory. The object's key needs to be
+  the id that's guaranteed to always be available.
+* The objects under used and free directories under disks, need to be empty
+  objects with their key being the id of the disk object under disks.
+```
+     list: nodes/$Node_context.node_id/Disks
+     list: nodes/$Node_context.node_id/Disks/Used
+     list: nodes/$Node_context.node_id/Disks/Free
+```
+
+== Impacted modules
+
+== Tendrl API impact:
+
+None
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+* Add a new function in pull_hardware_inventory to collect disk details.
+* Create a class Disk in persistence module.
+* Modify the manager to retrieve inventory details from tendrl-node-inventory
+  file and compare it with node_inventory list, If any changes then populate the
+  disk details in class instance and store it in etcd.
+* Add three different object in tendrl definitions for disks, used disks and
+  free disks.
+* Commands used to retrieve disk details are:
+```
+1. columns = 'NAME,FSTYPE,MOUNTPOINT,UUID,SIZE,ROTA
+   lsblk --all --bytes --noheadings --output='%s' --path --raw" % columns
+2. hwinfo --block (It will give disk, partitons, cdrom information. For disk
+   details disk and partitions are enough).
+```
+* Expected disk details output format form  hardware inventory list is:
+```
+{"disks": [{"disk_id": "",
+            "parent_id": "",
+            "disk_type": "",
+            "model": "",
+            "vendor": "",
+            "serial_no": "",
+            "device_name": "",
+            "sysfs_id": "",
+            "sysfs_busid": "",
+            "sysfs_device_link": "",
+            "driver_modules": "",
+            "driver": "",
+            "device_files": "",
+            "device_number": "",
+            "device": "",
+            "drive_status": "",
+            "rmversion": "",
+            "bios_id": "",
+            "geo_bios_edd": "",
+            "geo_bios_legacy": "",
+            "geo_logical": ""
+            "disk_kernel_name": ""
+            "parent_name": "",
+            "major_to_minor_no": "",
+            "fstype": "",
+            "mount_point": "",
+            "label": "",
+            "fsuuid": "",
+            "read_ahead": "",
+            "read_only": "",
+            "removable_device":"",
+            "size": "",
+            "state": "",
+            "owner": "",
+            "group": "",
+            "mode": "",
+            "alignement": "",
+            "min_io_size": "",
+            "optimal_io_size": "",
+            "phy_sector_size": "",
+            "log_sector_size": "",
+            "disk_type": "",
+            "ssd": "",
+            "scheduler_name": "",
+            "req_queue_size": "",
+            "discard_align_offset": "",
+            "discard_granularity": "",
+            "discard_max_bytes": "",
+            "discard_zeroes_data": "", ...]
+  "free": [disk_id,...],
+  "used": [disk_id,.....]...}
+```
+
+=== Tendrl/gluster_integration impact
+
+None
+
+=== Tendrl/ceph_integration impact
+
+None
+
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+None
+
+
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
+* "hwinfo" and "lsblk" to find the disk details.
+* lsblk is not sufficient to find all disk details, So we have to use hwinfo also.
+* From node_agent add a new function in pull_hardware_inventory to find disk
+  details and update those details into node_inventory list.
+* Add all possible disk details in hardware inventory list.
+```
+  (disk_id, device_name, internal_kernel_device_name, parent_id, parent_name,
+  maj_to_min_device_number, fs_type, mount_point, label, fs_uuid, read_ahead,
+  read_only, removable_device, model, size, state, owner, group, mode, alignement,
+  min_io_size, optimal_io_size, physical_sector_size, logical_sector_size, ssd,
+  scheduler_name, request_queue_size, disk_type, discard_alignment_offset,
+  discard_granularity, discard_max_bytes, discard_zeroes_data, vendor, serial_no,
+  firmware_version, sysfs_id, sysfs_busid, sysfs_device_link, driver, driver_modules,
+  device_files, device_number, bios_id, geometry_logical, geometry_bios_legacy).
+```
+* Create a class called  Disks  in persistence module to keep all disk details.
+* Modify the manager to retrieve inventory details from tendrl-node-inventory
+  file and compare it with node_inventory list, If any changes then populate the
+  disk details in class instance and store it in etcd.
+```
+- Populate all disk details in Disks class instance and store it in etcd.
+- store free disks id keys with empty object in etcd
+- store used disks id keys with empty object in etcd.
+```
+* Store disk class instance in etcd with key.
+* Create objects in tendrl definition for disks, used disks and free disks.
+* Disks structure in etcd:
+```
+/nodes/node_id/Disks
+   <disk_id1> # object with all the details for disk_id1
+   <disk_id2> # object with all the details for disk_id2
+
+/nodes/node_id/Disks/used
+   <disk_id1> # empty object
+
+/nodes/node_id/Disks/free
+   <disk_id2> #empty object
+```
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* git-hub issue: https://github.com/Tendrl/node_agent/issues/78
+
+== Dependencies
+
+Add hwinfo to retrieve disk details (pip package is not required for these two,
+It is used for system command). hwinfo is not available for rhel and epel, we
+have to build this.
+
+== Testing
+
+* Sanity check for flow.
+* Check all disk details are persisted successfully.
+
+== Documentation impact
+
+None
+
+== References
+
+None


GowthamShanmugam:
@shtripat @r0h4n @nnDarshan please review it

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,113 @@
+= Specification for adding disk details in node hardware inventory
```
nnDarshan, line 1: Please refer my comments from: https://github.com/Tendrl/specifications/pull/19#pullrequestreview-11780721 . More or less all the suggestions hold good here as well

nnDarshan (changes requested):

GowthamShanmugam:
@nnDarshan @shtripat @r0h4n @Tendrl/qe Please review it

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: i will change it

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Add disk details also as a part of node hardware inventory
+
+== Problem description
+
+Currently disk details are missing in node hardware inventory. Add disk details
+in node hardware inventory and disk details should contain device_name, fs_type,
+fs_uuid, model, mount_point, parent, size, name, disk_type, used, ssd, vendor,
+disk_id, node_id.
```
cybernth: There is an issue of disk names getting remapped when one of the disks are removed from existing inventory. @brainfunked was talking about some disk-id/fsid maintained by kernel can be used to maintain the uniqueness of the disk. This needs to figured out and added to this spec

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Api to get disk details v2/keys/nodes/node_id/Disks
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+Modify persistence and manager modules to find and store disks details.
```
cybernth: Any changes required in the tendrl definitions should be captured here

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Tendrl should show the current status of disk.
+
+== Proposed change
+
+* Modify manager module to find disk details and update those details into
+  node_inventory list.
+* Create a new class in persistence module to keep disk details.
+* Populate the object using node_inventory list and persist the object in etcd.
+
```
cybernth, line 18: More details should go in here like the command used and output as in the pull request https://github.com/Tendrl/specifications/pull/6

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,111 @@
...
+== Data model impact
+
+Add disk details in etcd with key nodes/node_id/Disks/device_name
+
+== Impacted modules
+
+== Tendrl API impact:
+
+Api to get disk details v2/keys/nodes/node_id/Disks
```
cybernth: Do we really need a separate API to get the disk details of the node?

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,111 @@
...
+None
+
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+We might have to monitor the status of the disk.
```
cybernth: How? If you think this should be addressed as part of another spec, raise github issues and link it here

cybernth (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: I will find it out and update in this specification file

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: i will add command and output in this spec file

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: I have added a field called "used", it will tell whether a disk is used or not.  I have mentioned this field as monitor here.

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: i will add it

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: I think no need, everything comes from node api. I will remove it

GowthamShanmugam:
@brainfunked I have checked disk id for formatted and un formatted disk, I can't find out any unique id for both disk.  A command ls /dev/disk/by-id  is giving id for all disks but it generated based on disk serial number, But this id is not in good format 


lrwxrwxrwx. 1 root root  9 Dec  9 17:07 ata-QEMU_DVD-ROM_QM00003 -> ../../sr0
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-name-centos_dhcp43--215-home -> ../../dm-2
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-name-centos_dhcp43--215-root -> ../../dm-0
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-name-centos_dhcp43--215-swap -> ../../dm-1
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-uuid-LVM-siZdB23ClU7bGFdOVaKmF9zbfrzHitPC7jfbdLR3L2ZXrZbfWxpmhsfAOZlzC8Rr -> ../../dm-0
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-uuid-LVM-siZdB23ClU7bGFdOVaKmF9zbfrzHitPCgY4PWjp2pkMPfW0YWs2zQ1ZOxdBDgcop -> ../../dm-2
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 dm-uuid-LVM-siZdB23ClU7bGFdOVaKmF9zbfrzHitPCIBcS9jnPs9Y3g7RmKXi2JeQXjWuKZmdh -> ../../dm-1
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 lvm-pv-uuid-rbjgZ0-WzdS-Jjjv-rSEN-Nldu-BCUI-at138w -> ../../vda2
lrwxrwxrwx. 1 root root  9 Dec  9 17:07 virtio-539e2d8b-86d9-4d43-a -> ../../vda
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 virtio-539e2d8b-86d9-4d43-a-part1 -> ../../vda1
lrwxrwxrwx. 1 root root 10 Dec  9 17:07 virtio-539e2d8b-86d9-4d43-a-part2 -> ../../vda2
lrwxrwxrwx. 1 root root  9 Dec  9 17:07 virtio-b8910937-0697-4030-9 -> ../../vdc
lrwxrwxrwx. 1 root root  9 Dec  9 17:07 virtio-ee4e2f4b-754f-4d6d-9 -> ../../vdb


I dont think it is usable

mkudlej:
@GowthamShanmugam I think using lsblk is better here. Please look at https://github.com/skyrings/skyring/blob/master/backend/salt/python/skyring/saltwrapper.py#L177

nnDarshan:
@mkudlej The initial plan was to use lsblk. But the concern we see is, when lsblk is used it doesn't  give UUID for a raw disk. UUID is available only for provisioned disks. If you are aware of any tool that can give UUID for all type of disks(raw and used) please let us know. 
NOTE: we cant use disk name to identify the disk as it might change after reboot.

cybernth:
@GowthamShanmugam suppose if you don't have any such id available, can you document an alternative approach to handle this?We need to make a decision and move on at the earliest.

nnDarshan:
If we don't  have such IDs available then, my suggestion would be to store used and raw disks separately. And use disk-uuid to identify used-disks and use disk-name to identify unused raw disks.something like:
 * nodes/node-id/disks/used/disk-uuid/  for used
 * nodes/node-id/disks/free/disk-name/ for raw disks.

We are mainly concerned about the rename of the disks that are being used, If we use UUID for disks that are being used this issue will be solved. And we are not bothered about the rename of unused disks, any how the disk details will be updated as part of hardware inventory sync. Also, if we go with the above approach, lsblk command would suffice.

GowthamShanmugam:
@nnDarshan am i modify the specification file as per your comment?

nnDarshan:
@GowthamShanmugam wait.
@nthomas-redhat @r0h4n @brainfunked Please provide your inputs

shtripat:
@nnDarshan your suggestion looks good to me but the only thing is the etcd dir names dont look good. May be the I would suggest rather apis as below
```
/nodes/<node-id>/used_disks

/nodes/<node-id>/free_disks
```
The only issue with this I see is that we would have keep removing entries from free_disks once they are moved to used.

cybernth:
@nnDarshan looks good to me. Please double check
- disk removal(free/used)
- free --> used transition
- used->free transition(OSD removed and formatted)
- new disk addition

@shtripat etcd path suggested by @nnDarshan (nodes/node-id/disks/used/disk-uuid/ ) looks better to me.

GowthamShanmugam:
@nnDarshan , @nthomas-redhat @shtripat please review it

shtripat:
@nthomas-redhat 
>> @shtripat etcd path suggested by @nnDarshan (nodes/node-id/disks/used/disk-uuid/ ) looks better to me.

My suggestion is almost matching with this. The only thing is I am merging words disks + used. so final apis could be like 
```
/nodes/<node-id>/used_disks/<disk-uuid>   for used ones
/nodes/<node-id>/free_disks/<disk-name>  for free ones
```

It reduces one level of apis in URL :)

GowthamShanmugam:
@nthomas-redhat @r0h4n @shtripat @nnDarshan @brainfunked review please

cybernth (approved):
LGTM

brainfunked:
Implemented in https://github.com/Tendrl/node_agent/issues/78

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,154 @@
...
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* Add Util-linux as a dependency.
+* Create a new class called in persistence module.
```
r0h4n: Create a new class called what?

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): TODO
+
+== Work Items
+
+* Add Util-linux as a dependency.
```
r0h4n: Is this available in epel and rhel6/7 and pip?

r0h4n (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: for rhel, util-linux,  pip i dont know

GowthamShanmugam: i will check

r0h4n: We need it available in PIP and epel

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,154 @@
...
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
```
r0h4n, line 148: There's a new dependency being added, please document impact @nthomas-redhat @TimothyAsir

r0h4n (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: New class nothing but Disk  in persistence to keep disk details

r0h4n: The statement is incomplete @GowthamShanmugam , "Create a new class called in persistence module"

shouldnt it be "Create a new class called **Disk** in persistence module" ?

r0h4n (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: Disk details are stored from inventory list to Disk class instance ,  and store instance in etcd

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: for epel it is available

GowthamShanmugam:
@r0h4n ok i will change it in specification file

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @r0h4n i am using lsblk command for find disk details.. do we really need this package... It is a default pakcage in linux right?

GowthamShanmugam: for pip it is not available

GowthamShanmugam: lsblk is a basic command in linux right.. it is available in all linux os... do we really need to specify this package?

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,154 @@
...
+       "model":      "model",
+       "mount_point": ["mountpoint", ...],
+       "name":       "name",
+       "parent":     "parentdevicename",
+       "size":       uint64,
+       "disk_type":  "type",
+       "used":       boolean,
+       "ssd":        boolean,
+       "vendor":     "string",
+       "disk_id":     "uuid"}, ...], ...}
```
brainfunked: How are we gathering unique disk ids?

specs/disk_details_for_hardware_inventory.adoc
brainfunked: ..and add the necessary dependencies to the RPM spec file.

specs/disk_details_for_hardware_inventory.adoc
brainfunked: This doesn't need to be in pip. We'll just need to execute system commands and parse their output.

specs/disk_details_for_hardware_inventory.adoc
brainfunked: Update the document with the class name please.

brainfunked (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: Unique id comes only for used disk

GowthamShanmugam: it is nothing but file system id

GowthamShanmugam:
@brainfunked it does not requires any dependency packages

julienlim (changes requested):
@brainfunked @GowthamShanmugam I'm not sure whose names I should be tagging in the comments as I can't easily tell who's the originator.

I couldn't see where the following is tracked:
-- serial number of the drive
-- firmware version
-- status and/or state information

Also, if a drive is a SSD but is using a SAS interface, what disk_type would it register as?  Similarly what would the ssd (boolean) reflect?  Do you really need both disk_type and ssd?

GowthamShanmugam:
@julienlim i will add serial number and firmware version . For firmware version i cant retrieve using lsblk, i have to use different tool for that. Status will comes under monitoring i feel. We can't say disk is up or down, if disk is removed then details of that particular disk are removed from etcd.  Disk utilization are covered in monitoring.

how i am using Disk_type means, i am only considering disks which is in type ['disk', 'part']. ssd is used particular disk is ssd or not.

@nnDarshan we need both right disk_type and ssd?

i am not sure what happens when SSd connected using SAS, @brainfunked i want your help for this

GowthamShanmugam:
@brainfunked @nthomas-redhat @r0h4n @anivargi @nnDarshan @shtripat As per anup comment, am i store disk details like :    
           for free disk nodes/node_id/Free_Disk
           for used disk nodes/node_id/Used_Disk

I want yours suggestion

GowthamShanmugam:
@julienlim  Partitions does not have serial number

GowthamShanmugam:
@r0h4n @nthomas-redhat @nnDarshan @brainfunked please review

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,156 @@
...
+  disk details in etcd.
+* Store free disk details with different key and used disk details with
+  different key.
+* Create objects in tendrl definition for used disks and free disks.
+* Commands used to retrieve disk details are:
+```
+1. columns = 'NAME,KNAME,FSTYPE,MOUNTPOINT,UUID,PARTUUID,MODEL,SIZE,TYPE,PKNAME,
+   VENDOR'
+   lsblk --all --bytes --noheadings --output='%s' --path --raw" % columns
+2. hdparm -I disk_name
```
brainfunked: How do you plan on parsing the hdparm output?

GowthamShanmugam: @brainfunked , Like how darshan implemented in other details, i have reffered same method, i just parsing out put  like spiting line by line and finding out serial_no and firmware version..
I have already implemented it
https://github.com/Tendrl/node_agent/pull/64

cybernth: Does hdparam comes by default in RHEL or you need any additional packages?

GowthamShanmugam: @nthomas-redhat it is not come by default , we have to specify. I have added hdparm in dependency.

cybernth: @GowthamShanmugam does that available in base RHEL channels?

GowthamShanmugam: @nthomas-redhat yes it is available

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+None
+
+== Data model impact
+
+* Add used disk details in etcd with key nodes/node_id/Used_disks.
+* Add free disk details in etcd with key nodes/node_id/Free_disks.
+* Add two different object in tendrl definitions for used_disk and free_disk.
+```
+     list: nodes/$Node_context.node_id/Free_disks
```
brainfunked: I would prefer the path suggested by @nnDarshan, where used and free are under disks. What I would do is to always add all the disks under the disks directory. Under used and free, I would simply specify the id of the disk whose details are in the disks directory. It's safer to keep the disk data in one place and just move references to it around. This also allows information for a specific disk to be queried directly. Without the references, one would have to scan both the free and used directories to find the appropriate disk and then query it. Always enable a direct query path for every object.

brainfunked (changes requested):
I think the SSD over SAS case @julienlim has brought up needs to be marked as an issue in the node agent repo and left open for now. I don't think we have the hardware resources readily available to be able to verify this, but perhaps we could initiate a conversation with the appropriate project teams to help us out with it.

In any case, @GowthamShanmugam please update the disk paths under nodes as I've requested and then this pull request is good to be merged.

specs/disk_details_for_hardware_inventory.adoc
cybernth: @GowthamShanmugam , hdparm won't work for scsi disks. You need to use sdparm command for that. So you need to use both wherever applicable. Also check the availability of sdparm in base RHEL. Document the package information and rpm spec changes required as well in the document

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked free disk does not have disk id, we cant store all disk under disks. What darshan told is free disk details should store  under /disks/free_disk using disk_name , used disk should store under /disks/used_disks using disk id

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @nthomas-redhat i will check and update it

specs/disk_details_for_hardware_inventory.adoc
brainfunked: Check out `hwinfo` and see if it provides the output we need. It is also able to dump the output to a file and compare that when invoked with --map to provide a list of changes since the last dump. If possible, replace hdparm with hwinfo. The hdparm output parsing is really fragile, even with the changes you've made in the code.

On the other hand, device names are unique on each node and they need to be unique only within the context of a specific node. While they can change, if we leverage hwinfo's map facility, we may be able to handle that situation without problems. Keep in mind that the device name will generally change only if the physical connections have been altered. The location of the device in the ATA bus will determine it's device name, to my knowledge.

GowthamShanmugam: hwinfo gives very hug data.. and extracting disk details and parsing the output from hwinfo os not better than lsblk i feel.. @nthomas-redhat @shtripat  any suggestions ?

GowthamShanmugam: @brainfunked  i don't see any rhel and epel package for hwifo

GowthamShanmugam: @brainfunked 
for hwinfo command  :  hwinfo --block

22: IDE 00.0: 10600 Disk                                        
  [Created at block.245]
  Unique ID: 3OOL.qPX1W_dGFo7
  Parent ID: w7Y8.HhbUszJC6y1
  SysFS ID: /class/block/sda
  SysFS BusID: 0:0:0:0
  SysFS Device Link: /devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0
  Hardware Class: disk
  Model: "SAMSUNG MZ7TE512"
  Vendor: "SAMSUNG"
  Device: "MZ7TE512"
  Revision: "6L0Q"
  Driver: "ahci", "sd"
  Driver Modules: "ahci"
  Device File: /dev/sda
  Device Files: /dev/sda, /dev/disk/by-id/ata-SAMSUNG_MZ7TE512HMHP-000L1_S1GJNSAG400778, /dev/disk/by-id/wwn-0x4d30445853885002
  Device Number: block 8:0-8:15
  BIOS id: 0x80
  Drive status: no medium
  Config Status: cfg=new, avail=yes, need=no, active=unknown
  Attached to: #17 (SATA controller)

23: None 00.0: 11300 Partition
  [Created at block.434]
  Unique ID: bdUI.SE1wIdpsiiC
  Parent ID: 3OOL.qPX1W_dGFo7
  SysFS ID: /class/block/sda/sda1
  Hardware Class: partition
  Model: "Partition"
  Device File: /dev/sda1
  Device Files: /dev/sda1, /dev/disk/by-id/ata-SAMSUNG_MZ7TE512HMHP-000L1_S1GJNSAG400778-part1, /dev/disk/by-id/wwn-0x4d30445853885002-part1, /dev/disk/by-uuid/dda9f15f-c5ec-4674-894a-d9ae57b8243c
  Config Status: cfg=new, avail=yes, need=no, active=unknown
  Attached to: #22 (Disk)



how we can parse this, it not giving in any json formate

GowthamShanmugam: And we can't identify which disk is used and which disk not used from this

GowthamShanmugam: @brainfunked i have run the command
hwinfo --block --save-config  /dev/sda2

output of this command is stored in vi /var/lib/hardware/udi/bdUI.SE1wIdpsiiC 

inside file : 

hwinfo.hwclasslist = '00004000000001'
hwinfo.sysfsid = '/class/block/sda/sda1'
hwinfo.unixdevicelist = { '/dev/sda1', '/dev/disk/by-id/ata-SAMSUNG_MZ7TE512HMHP-000L1_S1GJNSAG400778-part1', '/dev/disk/by-id/wwn-0x4d30445853885002-part1', '/dev/disk/by-uuid/dda9f15f-c5ec-4674-894a-d9ae57b8243c' }
hwinfo.unixdevice = '/dev/sda1'
hwinfo.baseclass = 275 (0x113)
hwinfo.active = 'unknown'
hwinfo.needed = 'no'
hwinfo.available = 'yes'
hwinfo.configured = 'new'
hwinfo.hwclass = 'partition'
hwinfo.model = 'Partition'
hwinfo.parentid = '3OOL.iLjwFstLQvC'
hwinfo.uniqueid = 'bdUI.SE1wIdpsiiC'

this the out put, here also line by line parsing is required

brainfunked: Check the hwinfo source code and use the parsing logic it itself uses to parse this file. You probably can't reuse the code, but you can certainly duplicate the parts of the logic that we need.

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+== Alternatives
+
+None
+
+== Data model impact
+
+* Add used disk details in etcd with key nodes/node_id/Used_disks.
+* Add free disk details in etcd with key nodes/node_id/Free_disks.
+* Add two different object in tendrl definitions for used_disk and free_disk.
```
brainfunked: There needs to be one object per disk under the `disks` directory. This object's key needs to be it's unique id. This unique id must always be present, whether the disk is free or in use. The `used` and `free` directories under disks need to simply contain objects that are the same id as the object in `disks`. The objects under `free` and `used` can be empty. The objects under `disks` should contain all the information about a disk, including it's partitions. We'll get the following structure:

> /nodes/<node_id>/disks
> |--> <disk_id1> # object with all the details for disk_id1
> |--> <disk_id2> # object with all the details for disk_id2
> |--> /used
> |--|--> <disk_id1> # empty object
> |--> /free
> |--|--> <disk_id2> #empty object

GowthamShanmugam:
@nthomas-redhat @shtripat @brainfunked please review this

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,178 @@
...
+device_name, fs_type, fs_uuid, model, mount_point, parent, size, name, disk_type,
+used, ssd, vendor, serial_no, firmware_version.
+
+== Use Cases
+
+Tendrl should show the current status of disk.
+
+== Proposed change
+
+* "hwinfo" and "lsblk" to find the disk details.
```
shtripat: I feel some of the details like classes etc should move to implementation section

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
+https://github.com/Tendrl/node_agent/issues/78
```
shtripat: Here we should talk about what classes, attributes, function etc to be written

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: done

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: done

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+* hdparm is also giving some disk details like serial number and firmware but it
+  works only for ATA disks types. For SCSI disk we have to use sdparm.
+
+== Data model impact
+
+* Add used disks id in etcd with key nodes/node_id/Disks/Used.
+* Add free disks id in etcd with key nodes/node_id/Disks/Free.
+* Add all disks details in etcd with key nodes/node_id/Disks
+* Add three different object in tendrl definitions for disks, used disks and
```
brainfunked: Not three different objects- one object per disk. I'd mentioned this in my previous comment yesterday, but it seems to have been lost.

There needs to be one object per disk, which contains all the information about the disk. This needs to be in the `disks` directory. Lower case, not capitalised. The object's key needs to be the id that's guaranteed to always be available.

The objects under `used` and `free` directories under `disks` (again, lower case, not capitalised), need to be empty objects with their key being the id of the disk object under `disks`.

Add validations to ensure that the same id doesn't fall under both `used` and `free`.

brainfunked (changes requested):

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked I have tried to say same :) i will change this paragraph

GowthamShanmugam:
@brainfunked review please

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,190 @@
...
+None
+
+=== Tendrl/common impact
+
+None
+
+=== Tendrl/node_agent impact
+
+* Add a new function in pull_hardware_inventory to collect disk details.
+* Create three different classes Disks, FreeDisks, UsedDisks in persistence
```
brainfunked: Disk, UnusedDisk, UsedDisk. Both the latter ones should inherit from Disk.

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,190 @@
...
+* Modify the manager to retrieve inventory details from tendrl-node-inventory
+  file and compare it with node_inventory list, If any changes then populate the
+  disk details in class instance and store it in etcd.
+* Add three different object in tendrl definitions for disks, used disks and
+  free disks.
+* Commands used to retrieve disk details are:
+```
+1. columns = 'NAME,FSTYPE,MOUNTPOINT,UUID,SIZE,ROTA
+   lsblk --all --bytes --noheadings --output='%s' --path --raw" % columns
+2. hwinfo --block --only disk_name
```
brainfunked: Are you not using the output to file functionality from hwinfo?

GowthamShanmugam: no i am not @nthomas-redhat told me no need to store the data in file and parse it directly

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked  i will modify it to inherit both classes

GowthamShanmugam:
@nthomas-redhat @brainfunked @shtripat review please

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -0,0 +1,225 @@
...
+  maj_to_min_device_number, fs_type, mount_point, label, fs_uuid, read_ahead,
+  read_only, removable_device, model, size, state, owner, group, mode, alignement,
+  min_io_size, optimal_io_size, physical_sector_size, logical_sector_size, ssd,
+  scheduler_name, request_queue_size, disk_type, discard_alignment_offset,
+  discard_granularity, discard_max_bytes, discard_zeroes_data, vendor, serial_no,
+  firmware_version, sysfs_id, sysfs_busid, sysfs_device_link, driver, driver_modules,
+  device_files, device_number, bios_id, geometry_logical, geometry_bios_legacy).
+```
+* Create three different classes called  Disks, FreeDisks, UsedDisks in persistence
+  module and inherit freeDisks and UsedDisks from Disks class.
```
shtripat: %s/freeDisks/FreeDisks/g

shtripat (commented):
LGTM. 1 minor comment.

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: i will change it

GowthamShanmugam:
@shtripat done

brainfunked (approved):
The Disk model needs to be defined as an object under the tendrl global namespace, along with node, cluster etc. However, that could be tackled in a separate specification.

brainfunked:
@anivargi Please check the data model and approve the specification if it meets your requirements.

GowthamShanmugam:
@r0h4n please review

GowthamShanmugam:
@r0h4n please review

shtripat (approved):

specs/disk_details_for_hardware_inventory.adoc
r0h4n: Any updates @nthomas-redhat ?

r0h4n (changes requested):

r0h4n (approved):

PR #24, opened 2016-12-08
Specification: specification to change the repository name changes

cybernth:
Signed-off-by: Nishanth Thomas <nthomas@nthomas.blr.redhat.com>

### DIFF
--- a/specs/repository_name_changes.adoc
+++ b/specs/repository_name_changes.adoc
@@ -18,13 +18,14 @@ None.
 == Proposed change
 
 Below repositories to be renamed:
-ceph_integration --> ceph-integration
-node_agent --> node-agent
-gluster_integration --> gluster-integration
-common --> commons
-tendrl-api --> api
-performance_monitoring --> performance-monitoring
-tendrl_frontend --> dashboard
+
+* `ceph_integration` -> `ceph-integration`
+* `node_agent` -> `node-agent`
+* `gluster_integration` -> `gluster-integration`
+* `common` -> `commons`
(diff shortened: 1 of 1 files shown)


specs/repository_name_changes.adoc
```diff
@@ -0,0 +1,151 @@
...
+Name of `Tendrl` github group itself is not affected by this change.
+
+
+== Problem description
+
+In Fedora, the RPM names usually match the upstream repository names. Currently for tendrl GitHub repositories are not aligned with rpm names. This would make it clearer to everyone that what code comes from where and help to maintain the overall packaging ecosystem consistency.
+
+== Use Cases
+
+None.
```
shtripat, line 16: Should we mention that the use case is rpms built for the tendrl projects?

cybernth: @shtripat , I don't think that is use case comes under the repo name change ??

cybernth:
@r0h4n @Tendrl/qe Please have a look. If looks good I would like to get this merged and make changes

specs/repository_name_changes.adoc
shtripat: My point was to mention about the intention behind the change in names. Its actually being done because rpm names should match the upstream project names. If thats correct, shouldn't we mention the same here?

Other than this looks good to me.

cybernth: Yes this is mentioned as part of the 'Problem description'

shtripat:
Looks good to me

r0h4n (approved):

PR #58, opened 2016-12-13
Use 'enabled' attribute to allow API requests

anivargi:
Tendrl API should use the 'enabled' attribute defined in the definitions yaml files to allow enabled requests only.

### DIFF
--- a/specs/api_use_enabled_attribute_from_definitions.adoc
+++ b/specs/api_use_enabled_attribute_from_definitions.adoc
@@ -1,0 +1,104 @@
+// vim: tw=79
+
+= Use 'enabled' attribute in objects to enable GET API requests
+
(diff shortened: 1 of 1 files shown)


specs/api_use_enabled_attribute_from_definitions.adoc
```diff
@@ -0,0 +1,104 @@
...
+need to have a GET API call. 
+
+== Use Cases
+
+Using the configuration from the definitions files help us to limit and control
+what the end user can access via the API.
+
+== Proposed change
+
+Tendrl API should use the 'enabled' attribute defined against each object to
```
r0h4n: "enabled" is defined for Flows as well, the flag should be respected while generating POST/PUT to these flows

r0h4n (changes requested):

r0h4n (approved):

Issue #54, opened 2016-12-12
Import cluster workflow


cybernth:
Currently to import a cluster, user expected to provide the all the nodes in the cluster as input. This will be inconvenient if the size of the cluster is large. Instead the usecase flow is expected to be modified as:
- select one of the node in cluster
- use the selected node to get the details of the cluster
- submit the request to all the nodes to import the cluster

brainfunked:
@nthomas-redhat I think it would be a good idea to detail out the import cluster workflow as a whole and incorporate these changes in the same specification.

brainfunked:
Linked to https://github.com/Tendrl/specifications/issues/56

brainfunked:
Depends on https://github.com/Tendrl/specifications/issues/46

cybernth:
@Tendrl/tendrl-core , @Tendrl/qe please review

cybernth:
@anivargi @brainfunked @r0h4n please review

brainfunked:
Can leverage https://github.com/Tendrl/specifications/issues/55

brainfunked:
Based on the updates (https://github.com/Tendrl/specifications/pull/100) now sent to the node agent namespaces specification (https://github.com/Tendrl/specifications/issues/87), the import cluster functionality would play out as follows:

Prerequisite:

* Node agents are installed on each of the nodes to be imported and can connect to etcd (auto-deployment of the node agents themselves is a provisioning task and will be tackled with the other provisioning workflows, in a separate specification).


Upon starting up, the node agent performs the following:

* Run the platform detection modules, stop at the first positive identification, gather the data and push it to etcd.
* Run the hardware and network inventory functionality and push the inventory to etcd.
* Run all the detection modules, gather all the data, create the `detected_cluster` object and populate etcd.
* Set the node in the correct state. For nodes where a storage system has been detected, the state would be `Import Pending`. For other 'clean' nodes, the state would be `Managed`.

This is how the import cluster workflow will be triggered from the API:

* A generic, hardcoded `ImportCluster` flow is defined globally in the master definition file. This flow requires the following parameters:
  - cluster_id
  - cluster_type
  - cluster_detection_id
  - cluster_name
  - cluster_nodes
  - Optionally, cluster_version
* `cluster_id` is generated by the API. At the time of the job creation, the appropriate cluster id based namespace is created under `/clusters`.
[...]

Issue #56, opened 2016-12-12
Import cluster UI


kamleshverma1:
We are going with dynamic generation and we have common form generator.
we are using this form generator for generate the form but in the import cluster case , we have different ux design , not common select box or textbox as we have in form generator.
so finally for import cluster we are not going ahead with dynamic generation for now. it will be static but we will use directive for table view.

brainfunked:
I'm renaming this specification to address it's scope properly. This needs to be linked to the import cluster related jira issues and the specification needs to address how you're planning on implementing the UI itself and at each step, link to the appropriate UX slide.

brainfunked:
Linked to https://github.com/Tendrl/specifications/issues/54

kamleshverma1:
Import cluster UX slide : 
https://docs.google.com/presentation/d/18jbUSbQbbcDBo4PI58q-uZbKTm0Ai0msePpsMfLio-k/edit#slide=id.g154cd53edc_0_0

mcarrano:
@kamleshverma1 Note that the UX link you pasted above is to an old design.  The latest/current design is posted to InVision here: https://redhat.invisionapp.com/share/R88EUSGJK

kamleshverma1:
Linked for Jira Tickets  : https://tendrl.atlassian.net/browse/TEN-171

kamleshverma1:
Linked for tendrl-frontend issue : https://github.com/Tendrl/tendrl_frontend/issues/63

julienlim:
I believe the Import Cluster JIRA user stories that this spec needs to link to are as follow:
* Import Ceph 2.x cluster (https://tendrl.atlassian.net/browse/TEN-4)
* Import Gluster 3.2 cluster (https://tendrl.atlassian.net/browse/TEN-3)

PR #60, opened 2016-12-13
Specification for the import cluster workflow

cybernth:
tendrl-bug-id: Tendrl/specifications#54

Signed-off-by: Nishanth Thomas <nthomas@redhat.com>

### DIFF
--- a/specs/import_cluster_workflow.adoc
+++ b/specs/import_cluster_workflow.adoc
@@ -1,0 +1,504 @@
+// vim: tw=79
+
+= Import cluster workflow
+
+The intent of import cluster workflow is to provide a mechanism to add an existing storage cluster(gluster, ceph etc) into tendrl for management.
+
+
+== Problem description
+
+This specification describes import cluster workflow which is a mechanism to add an existing storage cluster(gluster, ceph etc) into tendrl for management.
+This specification addresses the import use-cases of both ceph and gluster storage
+
+== Use Cases
+
+
+This addresses the use case of importing existing clusters(ceph. gluster etc.) into tendrl for management.
+
+The workflow is as follows:
+
+* Node Agents detect the capability of the storage nodes and pushes the details to central store.
+
+* The list of available nodes presented to the administrator via APIs
+  ** User selects one node in the cluster
+  ** Use the selected node to extract the details of the cluster
+  ** Submit import cluster request to nodes
+  ** On completion the cluster is expected to be imported into tendrl with all   objects(ceph pools, gluster volumes)
+
+* Sds integration components(ceph-integration, gluster-integration etc) pushes the cluster inventory to the central store
+
+== Proposed change
+
+
+Node Agents related changes:
+
+* Populate the node inventory(hardware, service etc) in the central store
+* Tendrl definitions - add necessary objects, atoms and flows for import cluster
+* Implement the atoms and flows for import cluster
+
+Tendrl API related changes:
+
+* API to get the cluster details. Input to this API would be:
+
+  ** Any node in the cluster in case of gluster
+  ** Any MON node in the cluster in case of ceph
+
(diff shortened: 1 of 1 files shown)


brainfunked:
As I'd commented in the issue, this needs to be a complete import cluster specification rather than a diff against the currently available implementation. The specification needs to include the whole import cluster functionality as it would be post this specification. Since we have already implemented the basic import cluster flow, in the spec, you can mention the additions that will be made, but the spec needs to be about the whole workflow.

cybernth:
@brainfunked  ack

julienlim (changes requested):
@nthomas-redhat and @brainfunked -- In reviewing the Import Cluster Workflow, I would like to raise some questions / concerns that are not covered in the spec (or at least I couldn't find them):

(1) How do you distinguish clusters created via TripleO vs. Heketi vs. gdeploy vs. manually vs. Tendrl vs. other?  This will impact tolerance and validations later on.  Do we persist an attribute for this?  

(2) Do we validate POC vs. Production deployment configuration?  Do we persist an attribute for this?  Is regularly validated (as new hosts or configs can change over time)?  This will affect validations in other places especially provisioning workflows post cluster import and for various Day 2 use cases.

(3) For clusters with CephFS, how do we note this?  We may need to provide some additional warnings for PG changes (as the pgcalc does not factor in CephFS today) and other places in the UI where utilization levels may not accurately depict the full picture of the cluster.

(4) Have we considered a validation framework / module in Tendrl much like the one used in TripleO (see https://github.com/rthallisey/clapper) that is modular and extensible that is used to perform pre-deployment, in-flight deployment, and post-deployment validations for deployment OpenStack and Ceph?  The reason is new validations will get added all the time especially as the support / compatibility matrix gets updated over time.  Also, it will get used by more than just the Import Cluster but just about every other use case (epics and user stories) in Tendrl.

cybernth:
> @nthomas-redhat and @brainfunked -- In reviewing the Import Cluster Workflow, I would like to raise some questions / concerns that are not covered in the spec (or at least I couldn't find them):
> 
> (1) How do you distinguish clusters created via TripleO vs. Heketi vs. gdeploy vs. manually vs. Tendrl vs. other? This will impact tolerance and validations later on. Do we persist an attribute for this?
> 

**Do we really need to? I think Tendrl should be agnostic of by what means the cluster is created. What Tendrl should care about is whether it is a valid configuration or not** 

> (2) Do we validate POC vs. Production deployment configuration? Do we persist an attribute for this? Is regularly validated (as new hosts or configs can change over time)? This will affect validations in other places especially provisioning workflows post cluster import and for various Day 2 use cases.
> 

**Support for POC configurations is something which needs to discussed in a different specification. As far as import is concerned , there should be an atom to validate the cluster getting imported is of valid configuration** 

> (3) For clusters with CephFS, how do we note this? We may need to provide some additional warnings for PG changes (as the pgcalc does not factor in CephFS today) and other places in the UI where utilization levels may not accurately depict the full picture of the cluster.
> 

**CephFS is something we need to worry about at this point?  Notifications on PG changes and utlization is not covered under this specification. Please note there is another spec(https://github.com/Tendrl/specifications/issues/56) being worked up on which covers the UI/UX side of the import.  PR is https://github.com/Tendrl/specifications/pull/64**  

> (4) Have we considered a validation framework / module in Tendrl much like the one used in TripleO (see https://github.com/rthallisey/clapper) that is modular and extensible that is used to perform pre-deployment, in-flight deployment, and post-deployment validations for deployment OpenStack and Ceph? The reason is new validations will get added all the time especially as the support / compatibility matrix gets updated over time. Also, it will get used by more than just the Import Cluster but just about every other use case (epics and user stories) in Tendrl.

**Tendrl follows a different model for validations. A flow has provisions to hook various pre and post validation atoms which does the job for us.**

PR #61, opened 2016-12-13
Specification: Centos CI utilization

dahorak:
This pull request adds centos_ci.adoc (specification for Centos CI utilization).

### DIFF
--- a/specs/centos_ci.adoc
+++ b/specs/centos_ci.adoc
@@ -2,28 +2,46 @@
 
 = CentOS CI
 
-CentOS CI should be configured to launch applicable set of tests as an
-acceptance tests for every new built rpm package.
-
 == Problem description
 
-None
+We need to make sure that new code doesn't break existing codebase and check
+that new features works as early as possible. To make this happen, CentOS CI
+will be configured to launch acceptance integration test run for every new
+nightly rpm build from tendrl nightly copr project. As set of implemented
+features in tendrl codebase grows, the set of acceptance tests is extended to
+cover new features.
 
 == Use Cases
 
-* Acceptance testing for every new rpm package.
+* Acceptance testing for every new build from tendrl copr nightly project.
+* The same set of tests should be run for stable builds from current tendrl
+  copr as well.
 
 == Proposed change
 
-* CentOS CI host provisioning system have to be utilized for testing machines
+* New copr tendrl project with nightly builds needs to be created. Right now
+  we have just copr project for
+  https://copr.fedorainfracloud.org/coprs/tendrl/tendrl/[stable tendrl buils],
+  but for the purposes of CI proposed here, we need another one with the latest
+  code.
+
+* When nightly build fails, no acceptance tests are executed and the build
+  needs to be fixed first.
+
+* Acceptance tests executed by CentOS CI system will work with rpm builds only.
+
+* CentOS CI host provisioning system will be utilized for testing machines
   provisioning.
 
 * Installation and configuration of all Tendrl components should be automated
-  for usage in CI (this contains also installation and configuration of all
-  additional required components including gluster and ceph).
+  via ansible playbooks/roles from
+  https://github.com/Tendrl/usmqe-setup[usmqe-setup] repository for usage in CI
+  (this contains also installation and configuration of all additional required
+  components including gluster and ceph).
 
 * Tendrl test suite should allow execution of subset of tests marked as
-  acceptance.
+  acceptance. http://doc.pytest.org/en/latest/example/markers.html[Pytest
+  marker] `@pytest.mark.acceptance` will be used to flag a test case as such.
 
 === Alternatives
 
@@ -35,7 +53,7 @@ None
 
 === Impacted Modules:
 
-None
+Every module needs to in good shape from rpmbuild point of view.
 
 ==== Tendrl API impact:
 
@@ -75,7 +93,8 @@ None
 
 === Developer impact:
 
-None
+When the nightly builds fail, build issues needs to be fixed first before CI
+system could run the acceptance tests.
 
 == Implementation:
 
@@ -86,6 +105,7 @@ Primary assignee:
 
 Other contributors:
   mkudlej
+  mbukatov
 
 === Work Items:
 
@@ -95,18 +115,18 @@ Other contributors:
 
 == Dependencies:
 
-TBD
+Nightly copr builds. Details of nightl copr build workflow should be specified.
 
 == Testing:
 
-None
+This is all about testing.
 
 == Documentation impact:
 
-None
+QE team needs to write documentation of the CI system when it's design is
+agreed on and finalized.
 
 == References:
 
 * https://ci.centos.org/view/tendrl/[CentOS CI - Tendrl board]
 * https://copr.fedorainfracloud.org/coprs/tendrl/tendrl/[Tendrl Copr Repository]
-
--- a/specs/centos_ci.adoc
+++ b/specs/centos_ci.adoc
@@ -3,7 +3,7 @@
 = CentOS CI
 
 CentOS CI should be configured to launch applicable set of tests as an
-acceptance tests for every new build.
+acceptance tests for every new built rpm package.
 
 == Problem description
 
@@ -11,7 +11,7 @@ None
 
 == Use Cases
 
-* Acceptance testing for new builds.
+* Acceptance testing for every new rpm package.
 
 == Proposed change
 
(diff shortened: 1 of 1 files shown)


specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
+// vim: tw=79
+
+= CentOS CI
+
+We need to make sure that new code doesn't break existing codebase and check
+that new features works as early as possible. This can be done by running
```
shtripat: %s/features works/features work/g

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+acceptance integration tests automatically, using CentOS CI infrastructure.
+
+== Problem description
+
+CentOS CI
+will be configured to launch acceptance integration test run for every new
+nightly rpm build from tendrl nightly copr project. As set of implemented
```
shtripat, line 13: Should we link to copr project here?

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+None
+
+=== Data model impact:
+
+None
+
+=== Impacted Modules:
+
+Every module needs to be in good shape from rpmbuild point of view.
```
shtripat, line 64: Can we list down what all should be run to make sure this is ready?

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+None
+
+=== Developer impact:
+
+When the nightly builds fail, build issues needs to be fixed first before CI
+system could run the acceptance tests.
+
+== Implementation:
+
```
shtripat, line 108: Add the specs Issue no here

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+* Jenkins job for installation of Tendrl and all other required components.
+* Jenkins job for launching subset of tests on prepared cluster.
+
+== Dependencies:
+
+Nightly copr builds. Details of nightl copr build workflow should be specified.
+
+== Testing:
+
+This is all about testing.
```
shtripat, line 130: I feel this is about testing the testing framework ;)

brainfunked:
Against https://github.com/Tendrl/specifications/issues/53

specs/centos_ci.adoc
```diff
@@ -2,28 +2,49 @@
...
 
-* Acceptance testing for every new rpm package.
+* Acceptance testing for every new build from tendrl copr nightly project.
+* The same set of tests should be run for stable builds from current tendrl
+  copr as well.
 
 == Proposed change
 
-* CentOS CI host provisioning system have to be utilized for testing machines
+* New copr tendrl project with nightly builds needs to be created. Right now
```
cybernth, line 25: why copr? Can we use the centos infra itself to build and host?

specs/centos_ci.adoc
```diff
@@ -2,28 +2,49 @@
...
+* Acceptance tests executed by CentOS CI system will work with rpm builds only.
+
+* CentOS CI host provisioning system will be utilized for testing machines
   provisioning.
 
 * Installation and configuration of all Tendrl components should be automated
-  for usage in CI (this contains also installation and configuration of all
-  additional required components including gluster and ceph).
+  via ansible playbooks/roles from
+  https://github.com/Tendrl/usmqe-setup[usmqe-setup] repository for usage in CI
```
cybernth, line 46: Need more details here rather than providing the link to the repository. This specification should provide details about:
1. How to access the centos infrstructure
2. How the nightly packages are built
3. How a test gets executed
4. Where to look for the the results

cybernth (changes requested):

specs/centos_ci.adoc
brainfunked: :+1:

specs/centos_ci.adoc
brainfunked: We should.

However, I'm wondering how this will actually play out. We need to be able to nightly test all the specifications that are in progress. This means that we need to set some way by which the packages can be built for specific branches nightly. These branches would be spread across multiple component, linked to from the specification issue. My thought is that we need to specify the branch name to be used in every repository across the project that contributes towards a specification. The name can be something as straightforward as `spec-$(issue_id)`.

In any case, would such builds be possible with copr? If not, we should be able to use the CentOS CI infrastructure itself to build artifacts and repositories and use them for individual specifications' nightly test runs.

Need inputs from @nthomas-redhat here.

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+== Use Cases
+
+* Acceptance testing for every new build from tendrl copr nightly project.
+* The same set of tests should be run for stable builds from current tendrl
+  copr as well.
+
+== Proposed change
+
+* New copr tendrl project with nightly builds needs to be created. Right now
```
brainfunked, line 25: Have added a comment above regarding this.

specs/centos_ci.adoc
```diff
@@ -0,0 +1,135 @@
...
+  we have just copr project for
+  https://copr.fedorainfracloud.org/coprs/tendrl/tendrl/[stable tendrl buils],
+  but for the purposes of CI proposed here, we need another one with the latest
+  code.
+
+* When nightly build fails, no acceptance tests are executed and the build
```
brainfunked, line 31: Indeed. This will put the responsibility of ensuring that the package spec files are kept up to date with each pull request onto the developers.

specs/centos_ci.adoc
brainfunked: We should in fact automate this as part of the build process itself. `rpmlint` output or some such needs to accompany every build.

brainfunked (changes requested):
Would like more details about how this would be implemented in Jenkins. It would have an impact on the coding practices such as topic branches, RPM spec file updates within each pull request etc. We'll probably need some standards and/or conventions to be set across the entire project to ensure that this works smoothly.

specs/centos_ci.adoc
cybernth: @brainfunked yes copr has this capability. You can configure nightly/per pull request based builds

specs/centos_ci.adoc
mbukatov: No particular reason for copr, we could use centos koji instance as well. This was just my initial proposal.

specs/centos_ci.adoc
mbukatov: > How the nightly packages are built

This is a question for the dev team I guess. From qe team point of view, I would be ok with any build process which is properly automated using either fedoraproject's copr or centos's koji so that it's possible to identify nighly rpm builds (via tags, generated yum repository) for each date.

mbukatov: > How a test gets executed

In the specification, I propose that the acceptation tests will be tagged via `acceptance`. Right now, there are no tests tagged with this marker.

mbukatov: > Where to look for the the results

Right now, we are not at the point when I could specify this in detail. I would expect that we have results available via centos jenkins instance and when we know the details, I could add it to the spec. The results will be presented via jenkins, which will read the data from xunit xml file generated by pytest.

specs/centos_ci.adoc
mbukatov: @shtripat I mean something bit different here: for the code in a nightly build branch to be tested via CI triggered acceptance tests, the builds needs to succeed. Without the packages, no tests would be executed. This is obvious, but I still wanted to write this down. The reason for this is that when some change breaks the build or is not complete so that the build would fail, fixing the build would be on the same level as fixing other issues, because without that, no acceptance tests would be run.

specs/centos_ci.adoc
mbukatov: What specification do you expect to be listed here?

specs/centos_ci.adoc
mbukatov: This is being discussed under https://github.com/Tendrl/specifications/issues/53 as well. As I noted there in https://github.com/Tendrl/specifications/issues/53#issuecomment-268336489 , I worry about number of concurrent, independent builds (one for each specification under development).

specs/centos_ci.adoc
shtripat: ok. So at least I feel we should mention to run a command before pushing the changes and make sure rpmbuild readiness is achieved.

specs/centos_ci.adoc
shtripat: There would an issue created for this spec. The same we need to add here.

specs/centos_ci.adoc
mkudlej: Ok, it is already mentioned here. https://github.com/Tendrl/specifications/issues/53

r0h4n:
@dahorak Any updates on this, or may I close this one?

Issue #62, opened 2016-12-13
Provide a way to expose the time series data in time-series db to the Tendrl/tendrl-api


anmolbabu:
* The performance-monitoring application should expose rest apis to enable access to the performance data in time-series db.
* The tendrl-api proxies to the apis exposed by tendrl-api and hence is agnostic to specifics of time-series db.

PR #20, opened 2016-12-06
Add a monitoring provisioning flow

anmolbabu:
Add a monitoring provisioning flow

tendrl-bug-id: Tendrl/specifications#42
Signed-off-by: anmolbabu <anmolbudugutta@gmail.com>

PR #63, opened 2016-12-13
Specification for exposing apis to access time-series data

anmolbabu:
tendrl-bug-id: Tendrl/specifications#62

### DIFF
--- a/specs/time_series_api.adoc
+++ b/specs/time_series_api.adoc
@@ -33,10 +33,10 @@ it can be co-resident with tendrl application but nothing necessiates this.
 Following are its sub-components based on the functionalities:
 
 * Api Layer : This layer will serve the following responsibilities.
-  ** Serve the time series data stored in time series database.
-  ** The apis internally use a plugin based approach for querying the
-     time-series performance data which eases the task of supporting multiple
-     time-series databases.
+** Serve the time series data stored in time series database.
+** The apis internally use a plugin based approach for querying the
+   time-series performance data which eases the task of supporting multiple
+   time-series databases.
 * Aggregator : This module aggregates atomic stats(ex: cluster utilization) to
   composite(system utilization) stats which will be required by UI for the
   dashboards.
@@ -57,36 +57,35 @@ data in the time series db if the monitoring stack is installed.
 
 * Tendrl/performance-monitoring will expose apis to access performance data in
   time-series db. This is achieved by:
-  ** Interface to time-series db is designed in the form of plugins specific to
-     a time-series db.
-    *** This means that to support a new time-series db, all it takes is the
-        addition of its specific interface as a plugin and doesn't require
-        touching any other part of the code.
-    *** The way the plugins work is :
-      **** The interfaces to time-series db are added under the directory
-           time_series_db/dbplugins of tendrl/performance-monitoring.
-      **** These plugins inherit from a common interface and hence all of them
-           follow a similar basic skeletal structure.
-      **** The plugin framework uses python's meta-class approach and they get
-           registered with the framework when everything in the plugin path is
-           imported but only the plugin configured in trendrl's conf will be
-           used to interface to the time-series db. The way this works is as
-           follows:
-            ***** A singleton manager looks for all/any plugins in directory
-                  time_series_db/dbplugins and imports them all.
-            ***** The moment they are imported the meta-class registers an
-                  instance of the plugin with itself.
-            ***** The singleton manager then stores the instance of required
-                  time-series db plugin which will be henceforth used for all
-                  interactions with the time-series db.
-            ***** Since the manager is a singleton, there is only one instance
-                  of it throughout the life of Tendrl/performance-monitoring
-                  application which means all this house-keeping is done only
-                  once during the lifecycle of the application.
+** Interface to time-series db is designed in the form of plugins specific to
+   a time-series db.
+*** This means that to support a new time-series db, all it takes is the
+    addition of its specific interface as a plugin and doesn't require touching
+    any other part of the code.
+*** The way the plugins work is :
+**** The interfaces to time-series db are added under the directory
+     time_series_db/dbplugins of tendrl/performance-monitoring.
+**** These plugins inherit from a common interface and hence all of them follow
+     a similar basic skeletal structure.
+**** The plugin framework uses python's meta-class approach and they get
+     registered with the framework when everything in the plugin path is
+     imported but only the plugin configured in trendrl's conf will be used to
+     interface to the time-series db. The way this works is as follows:
+***** A singleton manager looks for all/any plugins in directory
+      time_series_db/dbplugins and imports them all.
+***** The moment they are imported the meta-class registers an instance of the
+      plugin with itself.
+***** The singleton manager then stores the instance of required time-series db
+      plugin which will be henceforth used for all interactions with the
+      time-series db.
+***** Since the manager is a singleton, there is only one instance of it
+      throughout the life of Tendrl/performance-monitoring application which
+      means all this house-keeping is done only once during the lifecycle of
+      the application.
 * The following apis will be exposed by Tendrl/performance-monitoring:
-  ** An api to list the monitored resources of a particular entity
-  ** An api to get statistics of any monitored resource from list in above api
-     for a particular entity.
+** An api to list the monitored resources of a particular entity
+** An api to get statistics of any monitored resource from list in above api
+   for a particular entity.
 * The Tendrl/tendrl-api proxies to these apis and remains agnostic of the
   time-series db specifics.
 
--- a/specs/time_series_api.adoc
+++ b/specs/time_series_api.adoc
@@ -196,7 +196,11 @@ GET /monitoring/nodes/<node_id>/<resource_name>/stats?from=1997-07-16T19:20+01:0
 ----------
 
 ----
-Note: The time used in the above api needs to abide the iso8601 format
+Note:
+* The time used in the above api needs to abide the iso8601 format
+* Time zone conversions from input iso8061 format user specified timezone to
+  graphite server timezone for queries to graphite db will be handled
+  internally by the api.
 ----
 
 To get the latest stat use:
--- a/specs/time_series_api.adoc
+++ b/specs/time_series_api.adoc
@@ -107,6 +107,10 @@ custom collectd plugins rather than the collectd provided ones
 
 The tendrl-api interfaces to the apis mentioned above under the section
 "Proposed change" and detailed below under "Notifications/Monitoring impact".
+Making call's to the monitoring app is via a http client programatically. The
+tendlr-api will define its own url's, somewhat similar to what API's from the
+monitoring app. We need to take this approach since we need to have one point
+of interaction.
 
 ==== Notifications/Monitoring impact:
 
@@ -216,8 +220,38 @@ None
 === Security impact:
 
 The apis exposed by Tendrl/performance-monitoring although are only getters,
-need to be covered by a layer of authentication.It could even be as sample as
-a shared token by which its access is enabled only to Tendrl/tendrl-api.
+need to be covered by a layer of authentication. This is achieved as:
+
+* Flask-Login provides user session management for Flask. It handles the common
+  tasks of logging in, logging out, and remembering users’ sessions over
+  extended periods of time.
+* It will:
+** Store active user’s ID in the session, and lets log them in and out easily.
+** Handle the normally-tricky “remember me” functionality.
+** Help protect users’ sessions from being stolen by cookie thieves.
+* It does not:
+** Impose a particular database or other storage method. How users are loaded
+   is entirely upto flask user.
+** Handle permissions beyond “logged in or not.”
+** Handle user registration or account recovery.
+
+So in accordance with above, the approach we will be taking is:
+
+* Add a default user admin with a configurable password(admin by default).
+* The default login auth cookie is alive for 365 days unless user explicitly
+  logged out. But this cookie alive period is exposed for configuration by
+  flask as a field by name 'REMEMBER_COOKIE_DURATION'.
+
+----
+Note:
+
+* The configuration of admin password, and addition of any more possible users
+  will be handled by user management as admin will also be moved to users in
+  etcd once tendrl has user-management.
+* The configuration of 'REMEMBER_COOKIE_DURATION' field will be exposed for
+  configuration via an api that handles the similar case for tendrl-api as both
+  the apis can share a common configurable field in etcd.
+----
 
 === Other end user impact:
 
--- a/specs/time_series_api.adoc
+++ b/specs/time_series_api.adoc
@@ -251,6 +251,8 @@ Note:
 * The configuration of 'REMEMBER_COOKIE_DURATION' field will be exposed for
   configuration via an api that handles the similar case for tendrl-api as both
   the apis can share a common configurable field in etcd.
+* To track the authentication part a separate spec will be raised. And that
+  spec will fix https://github.com/Tendrl/specifications/issues/89
 ----
 
 === Other end user impact:
--- a/specs/time_series_api.adoc
+++ b/specs/time_series_api.adoc
@@ -1,0 +1,287 @@
+// vim: tw=79
+
+= Expose a way for tendrl-api to expose data from time series db
+
+== Introduction
+
+A small introduction to the components part of this spec
+(borrowed from Tendrl/documentation):
+
+=== Tendr/tendrl-api
+
+* Frontend/northbound API.
+* This is the integration point for any external management system/frontend
+  with Tendrl, including the Tendrl UI.
+* The App has the necessary logic to understand the state definitions stored in
+  the Central Store and present them over it’s API.
+* The App has the necessary logic to be able to invoke Operations through the
+  Central Store’s Job Queue implementation.
+* The App is completely stateless. All the state is stored in the Central
+  Store. The App essentially acts as a translation proxy between an external
+  system (including Tendrl UI) and the Central Store.
+* The App is intended to be primarily used for systems integration. As such,
+  the API is completely generic and self descriptive. The intended use case is
+  that any system wishing to leverage Tendrl uses the API’s descriptive
+  responses to construct targeted queries for a specific cluster instances
+  being managed by Tendrl.
+
+=== Tendrl Performance (performance_monitoring)
+
+It is a separate service that runs outside tendrl core stack and it relies on
+the state information in tendrl’s central store for its functioning. Typically
+it can be co-resident with tendrl application but nothing necessiates this.
+Following are its sub-components based on the functionalities:
+
+* Api Layer : This layer will serve the following responsibilities.
+  ** Serve the time series data stored in time series database.
+  ** The apis internally use a plugin based approach for querying the
+     time-series performance data which eases the task of supporting multiple
+     time-series databases.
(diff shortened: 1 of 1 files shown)


anmolbabu:
@anivargi I have documented from monitoring perspective could you please add specifics from the perspective of tendrl-api

@anivargi @shtripat @nthomas-redhat @r0h4n @brainfunked please review.

TODO: Create issues on related components and link it in spec

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+      [5.50091681946991, 1478495700],....
+    ]
+  }
+]
+----------
+
+The above response gives the statistics for last 24 hours corresponding to the
+'resource_name' passed to it. If a time range is passed as optional arguements
+to the above api, then the api responds with the statistics for that time
+range.
```
shtripat, line 190: Can you add a sample for time range based invocation as well here?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+* time_series_db_server - ip of the server where the time-series db resides.
+* time_series_db_port - port on which time series db can be acceseed for read.
+* performance_server_interface - the network interface only on which the
+                                 Tendrl/performance_monitoring serves
+----
+
+=== Developer impact:
+
+This has extensive impact on 2 modules tendrl-api and performance-monitoring
+and hence warrants a very high collaboration between the implementers of the
```
shtripat: %s/implementers/owners/g

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+and hence warrants a very high collaboration between the owners of the
+2 modules
+
+== Implementation:
+
+
+=== Assignee(s):
+
+
+Who is leading the writing of the code? Or is this a spec where you're
```
shtripat: I feel line 253-257 is not required. Its retained from sample I feel. Please remove.

shtripat (commented):
LGTM other than two minor comments I have.

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+* Tendrl/performance-monitoring will expose apis to access performance data in
+  time-series db. This is achieved by:
+  ** Interface to time-series db is designed in the form of plugins specific to
+     a time-series db.
+    *** This means that to support a new time-series db, all it takes is the
+        addition of its apecic interface as a plugin and doesn't require
+        touching any other part of the code.
+    *** The way the plugins work is :
+      **** The interfaces to time-series db are added under the directory
+           tendrl/performance-monitoring/time_series_db/dbplugins of
```
r0h4n: If these are etcd directories, please use "_tendrl/" instead of "tendrl/", Etcd hides directories with a leading "_" and we need to use it

r0h4n (changes requested):

specs/time_series_api.adoc
anmolbabu: No it is location in tendrl/performance-monitoring code-base where the specific time-series db handlers need to be added

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+                  imports them all.
+            ***** The moment they are imported the meta-class registers an
+                  instance of the plugin with itself.
+            ***** The singleton manager then stores the instance of required
+                  time-series db plugin which will be henceforth used for all
+                  interactions with the time-series db.
+            ***** Since the manager is a singleton, there is only one instance
+                  of it throughout the life of Tendrl/performance-monitoring
+                  application which means all this house-keeping is done only
+                  once during the lifecycle of the application.
```
cybernth: Line 58-87, does this implementaion covered as part of this spec?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+  {
+    "target": "{graphite_prefix.host_name.metric_name}",
+    "datapoints": [
+      [5.50091681946991, 1478495700],....
+    ]
+  }
+]
+----------
+
+The above response gives the statistics for last 24 hours corresponding to the
```
cybernth, line 187: is this the latest value or 24 hours value?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+
+None
+
+==== Sds integration impact:
+
+None
+
+=== Security impact:
+
+The apis exposed by Tendrl/performance-monitoring although are only getters,
```
cybernth, line 226: Please add the details here

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+in its specific collection/table named by the fully qualified resource name
+and this is very specific to the way collectd's plugins send metrics to time
+series db and hence not much control can be exercised here unless we use
+custom collectd plugins rather than the collectd provided ones
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+The tendrl-api interfaces to the apis mentioned above under the section
```
cybernth, line 108: I would like see more details on how the tendrl-api to performance-monitoring interaction happens

specs/time_series_api.adoc
```diff
@@ -0,0 +1,287 @@
...
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+Some of the config options as below need to be made available:
+
+----
+[time_series]
```
cybernth: Is this something to be covered as mentioned in  -  https://github.com/Tendrl/specifications/pull/11

cybernth (changes requested):

specs/time_series_api.adoc
anmolbabu: This is 24 hours value :) I cut short the sample because it was too huge and symbolically added ',.... ' notation

specs/time_series_api.adoc
anmolbabu: Yes I was adding to the template and while doing so I forgot to remove this :)

specs/time_series_api.adoc
anmolbabu: This spec talks about apis that fetch time series data. So I added more details like how the apis actually fetch the details i.e, the api talks to a manager that takes the responsibility of finding the correct plugins that actually talks to time-series db

specs/time_series_api.adoc
cybernth: How to get the latest value(single last updated )?

specs/time_series_api.adoc
anmolbabu: @anivargi Can you please provide your inputs here

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+  time-ranges.
+
+== Proposed change
+
+* Tendrl/performance-monitoring will expose apis to access performance data in
+  time-series db. This is achieved by:
+  ** Interface to time-series db is designed in the form of plugins specific to
+     a time-series db.
+    *** This means that to support a new time-series db, all it takes is the
+        addition of its apecic interface as a plugin and doesn't require
```
brainfunked: "apecic"?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+== Use Cases
+
+* If monitoring stack is installed by user, the user should be able to seamlessly
+  access tendrl maintained time series performance data over specified
+  time-ranges.
+
+== Proposed change
+
+* Tendrl/performance-monitoring will expose apis to access performance data in
```
brainfunked, line 58: General tip: Don't indent the nested lists.

specs/time_series_api.adoc
brainfunked: In that case, omit the tendrl/performance-monitoring prefix from the directory name, since you've mentioned the repository name at the end of the sentence anyway.

specs/time_series_api.adoc
brainfunked: :+1:

specs/time_series_api.adoc
anmolbabu: like the time option, there needs to be one more option that internally does the graphite o/p parsing like we used to do before I'll add a sample for that as well

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+The api list to be exposed will grow with time. Currently, the following apis
+can be considered:
+
+To explore the resources that are monitored, we can use the api described
+below.
+
+Sample Request
+
+----------
+GET /monitoring/nodes/<node_name>/monitored_types
```
brainfunked: The node_name will need to be understood by the client making this API call somehow. Aren't the main APIs using node ids? The clients should not have to translate. All our APIs must use the same identifiers for objects.

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+    "target": "{graphite_prefix.host_name.metric_name}",
+    "datapoints": [
+      [5.50091681946991, 1478495700],....
+    ]
+  }
+]
+----------
+
+The above response gives the statistics for last 24 hours corresponding to the
+'resource_name' passed to it. If a time range is passed as optional arguements
```
brainfunked, line 188: There would be a maximum time range for which the data is stored in the time series db. Even if this isn't configurable right now, this value needs to be made known somehow.

brainfunked: And speaking of which, what's the way to change the duration for which the data would be stored in the db? Can this be done on the fly while the system is still pushing updates to the time series db?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+None
+
+=== Other deployer impact:
+
+Some of the config options as below need to be made available:
+
+----
+[time_series]
+time_series_db = graphite
```
brainfunked: Do we really need to expose this option? Do we even support any db other than whisper?

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+time_series_db_server = 0.0.0.0
+time_series_db_port = 80
```
brainfunked: Does the API run as root? It must not.

specs/time_series_api.adoc
```diff
@@ -0,0 +1,285 @@
...
+If more than one person is working on the implementation, please designate the
+primary author and contact.
+
+Primary assignee:
+  performance_monitoring part: anmolbabu
+  tendrl-api part: Anup(anivargi)
+
+=== Work Items:
+
+* https://github.com/Tendrl/performance_monitoring/issues/7
```
brainfunked, line 291: @anivargi please create an issue on the api repository and link it here.

brainfunked (changes requested):
The document needs integration details from @anivargi as has already been pointed out. The point I'd like to be considered when adding these details is that the interface must appear to be uniform. To the client, it should not be apparent that there are in fact two separate APIs being accessed to provide this functionality.

specs/time_series_api.adoc
anivargi: @nthomas-redhat Making call's to the monitoring app via a http client programatically. The API will define its own url's, somewhat similar to what API;s from the monitoring app. We need to take this approach since we need to have one point of interaction.

specs/time_series_api.adoc
anmolbabu: Ok

specs/time_series_api.adoc
anmolbabu: Ok

specs/time_series_api.adoc
anivargi: @brainfunked https://github.com/Tendrl/tendrl-api/issues/43 
I will write a specification before we start implementing this.

specs/time_series_api.adoc
anmolbabu: Ok I'll remove it from here. Its anyway covered in https://github.com/Tendrl/specifications/pull/20 in lines 286 to 297

specs/time_series_api.adoc
anmolbabu: specific :)

specs/time_series_api.adoc
anmolbabu: Basically, the problem is that, the stats are stored in graphite with node names and it is the collectd plugin that decides the name to be associated with a metric its like chosen prefix.node_name.resource_type.stat_type . If only we override the custom collectd plugins and write our own plugins, we'll be able to customize this name. But anyway instead of the api user having to do the mapping, the performance_monitoring application can as well do it using details in etcd. I'll modify the spec to reflect this

specs/time_series_api.adoc
anmolbabu: I have not thought of this before. I'll explore if there's a way to do this

specs/time_series_api.adoc
anmolbabu: The plan is to be able to support multiple time series dbs although there will be only 1 officially recommended. Any way its pluggable so the moment we add a plugin to different db, there'll be 2 types of dbs of which any one can be used. So, we have this field.

specs/time_series_api.adoc
cybernth: @anmolbabu @anivargi , Please add these details to the doc

specs/time_series_api.adoc
cybernth: Please add these details to the doc

specs/time_series_api.adoc
anmolbabu: That's not actually a necessity. I just tried running flask the tool that we use for apis as non-root, non-sudo user and it works. And the graphite querying part also can work as any user its just a http get.

specs/time_series_api.adoc
anmolbabu: Ok. I'll update this soon. Looking through details on flask docs.

specs/time_series_api.adoc
anmolbabu: ok

specs/time_series_api.adoc
shtripat: Does graphite provide some sort of archiving mechanism and data which is older than say 2 yrs goes to archive??

specs/time_series_api.adoc
anmolbabu: ok

specs/time_series_api.adoc
anmolbabu: Done added

specs/time_series_api.adoc
anmolbabu: @shtripat I think I have seen something related to this. I'll explore more on this.
@shtripat @brainfunked can detailing on this go into a separate spec? As this spec is more about the procedure employed to access time series data and hence mainly exposing apis to do the same. I'll create a bug to track that and add a mention of that here.

specs/time_series_api.adoc
```diff
@@ -107,6 +107,10 @@ custom collectd plugins rather than the collectd provided ones
 
 The tendrl-api interfaces to the apis mentioned above under the section
 "Proposed change" and detailed below under "Notifications/Monitoring impact".
+Making call's to the monitoring app is via a http client programatically. The
+tendlr-api will define its own url's, somewhat similar to what API's from the
+monitoring app. We need to take this approach since we need to have one point
+of interaction.
```
cybernth, line 113: I  guess what you mentioned here was the APIs exposed by monitoring app. Its good to mention them as well here becuse, those are the one's exposed to the end user

specs/time_series_api.adoc
```diff
@@ -216,8 +220,38 @@ None
...
+** Handle permissions beyond “logged in or not.”
+** Handle user registration or account recovery.
+
+So in accordance with above, the approach we will be taking is:
+
+* Add a default user admin with a configurable password(admin by default).
+* The default login auth cookie is alive for 365 days unless user explicitly
+  logged out. But this cookie alive period is exposed for configuration by
+  flask as a field by name 'REMEMBER_COOKIE_DURATION'.
+
```
cybernth, line 248: Does flask support token based auth?

specs/time_series_api.adoc
anmolbabu: @anivargi Could you please help me add it

specs/time_series_api.adoc
anmolbabu: @nthomas-redhat As we discussed offline, this needs to be inline with the approach taken by tendrl-core api. So, I'll raise a different bug and correspondingly a different spec to address this.

specs/time_series_api.adoc
shtripat: @anmolbabu I am ok fro different spec

specs/time_series_api.adoc
```diff
@@ -0,0 +1,305 @@
...
+The above response gives the statistics for last 24 hours corresponding to the
+'resource_name' passed to it. If a time range is passed as optional arguements
+to the above api, then the api responds with the statistics for that time
+range.
+
+To get the statistics for a given time-range:
+
+----------
+GET /monitoring/nodes/<node_id>/<resource_name>/stats?from=1997-07-16T19:20+01:00&to=1997-07-18T19:20+01:00
```
brainfunked, line 195: What about time zones?

anmolbabu: The time format employed is ISO 8601. I think this is standard followed by tendrl-api as well

So in case of 1997-07-18T19:20+01:00, +01:00 is offset from UTC time so with UTC as base time and offset mentioned the problem of time-zones is overcome

@anivargi could you please add if I am missing something

specs/time_series_api.adoc
anmolbabu: As discussed offline I have added a reference to github issue that will address the auth part of monitoring stack

specs/time_series_api.adoc
brainfunked: @anmolbabu yes, but is it possible for queries to be time zone specific and will they be resolved automatically and correctly or do we need to implement some UTC based time zone conversion to resolve such queries?

cybernth (approved):

cybernth:
Looks good to me

shtripat (approved):
LGTM. One question from @brainfunked  need to be answered.

specs/time_series_api.adoc
anmolbabu: @brainfunked , The conversion needs to be implemented by us.

specs/time_series_api.adoc
anmolbabu: Done

specs/time_series_api.adoc
anmolbabu: Raised an issue to track this : https://github.com/Tendrl/specifications/issues/99

specs/time_series_api.adoc
anmolbabu: The issue is @ https://github.com/Tendrl/specifications/issues/89

r0h4n (approved):

specs/time_series_api.adoc
r0h4n: - Always use aware datetime object, i.e. with timezone information. That makes sure you can compare them directly (aware and unaware datetime objects are not comparable) and will return them correctly to users. Leverage "pytz" to have timezone objects.
- Use ISO 8601 as input and output string format. Use datetime.datetime.isoformat() to return timestamps as string formatted using that format, which includes the timezone information.

More here: https://julien.danjou.info/blog/2015/python-and-timezones

Issue #55, opened 2016-12-12
Updates for in flight operations


cybernth:
Updates include details for every atom executed, including timestamps, actual invocation (cli, api etc.), output etc.

r0h4n:
Is there a spec for this yet?

cybernth:
@r0h4n , not yet

brainfunked:
@nthomas-redhat Do refer to the review comment https://github.com/Tendrl/specifications/pull/94#pullrequestreview-14123541 on the logging specification as that should provide some inputs that could be used to construct this specification.

brainfunked:
Each operation should be created as directory, with the id of the operation as the key. Keys must be created [in order inside this directory](https://github.com/coreos/etcd/blob/master/Documentation/v2/api.md#atomically-creating-in-order-keys). The first key inside the directory must always be a dump of the job parameters and the contexts (tendrl, node). Subsequent updates to the operation must be posted as individual keys created in the directory. Each such object must be a properly tagged message, based on https://github.com/Tendrl/specifications/pull/94. Any node agent specific decisions (such as a job being picked up, distributed etc.) must also be stored as objects in the operation directory, as an update (these need not be displayed as a notification to the administrator).

The tagged messages would contain the appropriate metadata to enable the API to filter out only the notifications to be sent as job updates. The API would need to accept the key from which all subsequent updates must be returned. The absence of the key would indicate that all the updates regarding the operation must be returned. This feature is useful to deliver only the new messages to the client (UI, CLI etc.), for a tracked operation.

brainfunked:
This specification will be impacted by the comments on the logging specification pull request at https://github.com/Tendrl/specifications/pull/94#pullrequestreview-17399508. @GowthamShanmugam please update this spec accordingly.

brainfunked:
Given that the logging implementation specification, https://github.com/Tendrl/specifications/issues/28, is a superset of this specification and the pull request for it addresses all the requirements in this specification, https://github.com/Tendrl/specifications/pull/94, I'm copying the details from this specification over there and closing this one.

PR #94, opened 2016-12-19
Specification file for logging implementation

GowthamShanmugam:

tendrl-bug-id: Tendrl#28

### DIFF
--- a/specs/logging_implementation.adoc
+++ b/specs/logging_implementation.adoc
@@ -1,0 +1,327 @@
+= Generic message mechanism for logs and alerts in Tendrl"
+
+Logs and alerts need to be treated as messages emitted by Tendrl.
+
+== Problem description
+* The logs, alerts and operations updates in Tendrl are handled differently
+  and there is no standard mechanism which makes sure no internal details are
+  exposed as part of alerts/notifications sent to end user.
+* Ideally logs messages, operation updates and alerts are messages which
+  are emitted from an application.
+
+== Use Cases
+
+Logs and alerts need to be treated as messages.
+
+== Proposed change
+* Logs, alerts and operations need to be treated as messages emitted by Tendrl modules.
+* A schema needs to be defined for emitted messages.
+* Message priorities, types etc. need to be defined at the time of emission.
+* The integration/bridge applications must pass the message as per schema.
+* For messages that serve as notifications to the administrator (or alerts),
+  exact error message must be hardcoded by the developers. No backtraces or
+  system generated error messages should be sent as notifications to the
+  administrator.
+* Notifications sent to the administrator should provide sufficient context to
+  enable debugging of the problem via investigation into the log files.
+* Notifications and alerts need to be delivered to the node agent alert sockets.
+* Common module needs to implement a logging library that hides all the implementation
+  details and configuration options from any of the components using the library.
+  Components should be able to blindly emit messages to the Message Handler module
+  in Tendrl and it should be able to deliver them to either syslog, log files,
+  alert sockets etc. depending upon the metadata and configuration.
+* Syslog must be the default logging configuration. Logging to individual
+  configuration files should be thoroughly discouraged.
+
+== Alternatives
+
+* None
+
+== Data model impact
+
+* Store the operation details in etcd:
+```
+1. Operation messages with inorder_id
+    key: queue/{job_id_id}/messages/{in_order_id}
+
+    value: {
+       message_id : "uuid",
+       priority: "",
+       timestamp: time_now,
+       publisher: "node-agent/ceph-integration......" (particular module name)
+       "request_id": "",
+       "node_id": "",
+       "cluster_id": "", (optional for node related job)
+       "flow_id": "",
+       "parent_id", (optional)
+       payload: {
+          "message": ""
+       }
+    }
+
+2. All messages are sent to syslog, only notice and above are stored in etcd
+    - all the messages are stored in  /messages/{message_id}
+    - if message have node_id then it stored in /nodes/{node_id}/messages/{message_id}
+    - if the message have cluster_id then it stored /clusters/{cluster_id}/messages/{message_id}
+
+  value:{
+      message_id : "uuid",
+      priority: "",
+      timestamp: time_now,
+      publisher: "node_agent/ceph_integration......" (particular module name)
+      "node_id": "",
+      "cluster_id": "",
+      payload: {
+         "message": """
+         "exception_traceback": "" (created automatically by exception message class
+                                    only for exception messages)
+         "exception_type": "" (created automatically only for exception messages)
+         "exception": "" (Only for exception messages)
+      }
+   }
+```
+
+== Impacted modules
+
+=== Tendrl/common impact
+* Create a new class called "Message" in common.
+```
+structure of Message class
+    "message_id": <string, assigned by message handler>,
+    "priority": <string, defined by the sender>,
+    "timestamp": <string, defined by the sender>,
+    "publisher": <string, defined by the sender>,(it will tell message from which module,
+    used to find particular log configuration).
+    "payload": <json serialized dict, defined by the sender>
+    (payload will change for log and operation other details are common)
+```
+* The Message object itself should have validations to ensure that it's well formed.
+  The validator should check the object contain all the necessary information are populated
+  correctly. If the object is invalid then it create a new message object and store the
+  original object under payload in newly created object and set the priority as warning
+  and return the newly created object. If the object is valid then it return same object.
+* When the message object gets populated then it is passed to event class.
+* Create a new class called Event. Event class should emit a Message object and put that
+  into node_agent socket.
+```
+    Use of event class:
+      - Event class will takecare of connect node socket and put the message object
+        into it.
+            s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            s.connect(socket_path)
+            s.send(message_object)
+            s.shutdown(socket.SHUT_RDWR)
+```
+* Socket path is from configuration file (node-agent.conf.yaml).
+* Create a ExceptionMessage class to handle the exception messages.
+* ExceptionMessage class is a sub-class of Message class
+* ExceptionMessage class will find the exception traceback and format the treacback in
+  well defined formate.
+* ExceptionMessage class also converted to Message class before log and stored into
+  etcd.
+```
+ Mantatory paramerters for ExceptionMessage class is
+     - priority
+     - publisher
+     - payload:
+         - message
+         - exception
+```
+=== Tendrl/node_agent impact
+* Add job id, flow id, parent id in tendrl context.
+* Create message handler in node_agent to handle the incoming message objects from socket.
+```
+ Use of message handler:
+    - Configure and run the node agent socket for ever.
+    - Message handler will read the socket, and take the message object from socket.
+    - It will pass the message object to logger module.
+```
+* Run socket in node_agent and make it keep listen.
+```
+Configuring and run the socket by message handler:
+  - In node-agent.conf.yaml add socket address and port for configuring socket.
+      socket_path="some_socket_path"
+  - In manager.py Create an object for message_handler.
+  - Message_handler will configure the socket using socket_path.
+  - Create a function called start in message handler to start the socket.
+  - From manager.py with the help of message_handler object call the start function to
+    listen the socket.
+  - Create a stop function in message_handler to stop the socket.
+```
+* When the object is received by message handler, it will passed to logger module.
+* Create a logger module in node agent to handle the message object which is passed
+  from message_handler.
+* Create a new function called handle in logger module to process message object.
+```
+- All the messages are stored in etcd.
+       - all the messages are stored in /Messages/{message_id}
+       - if message have node_id then it stored in /nodes/{node_id}/Messages/{message_id}
+       - if the message have cluster_id then it stored /clusters/{cluster_id}/Messages/{message_id}
+
+ - If the message have "job_id" then it is considered as operation, store message
+   in operation directory in etcd with sequential key.
+   operational messages stored in queue/{job_id}/messages/{in_order_id}.
+
+      For generate a sequential key in a directory
+         client.write("queue/{job_id}/messages/", message_object, append=True)
+         It will store the message_object with sequence of keys
+
+- Store the all log messages via syslog.
+```
+
+=== Tendrl/gluster_integration impact
+* Import a Message class from common.
+* Prepare a well defined json and pass it to message class.
+* Import a Event class from common and pass the message object to it.
+
+=== Tendrl/ceph_integration impact
+
+* Import a Message class from common.
+* Prepare a well defined json and pass it to message class.
+* Import a Event class from common and pass the message object to it.
+
+=== Tendrl/alerting impact
+* Import a Message class from common.
+* Prepare a well defined json and pass it to message class.
+* Import a Event class from common and pass the message object to it.
+
+=== Tendrl/performance_monitoring impact
+
+* Import a Message class from common.
+* Prepare a well defined json and pass it to message class.
+* Import a Event class from common and pass the message object to it.
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+The collectd plugin to handle collectd detected threshold breaches should be
+customized to form the json in well defined format and put that in message handler
+socket.
+
+== Other end user impact
+
+None
+
+== Performance Impact
+
+None
+
+== Other deployer impact
+
+None
+
+== Developer impact
+
+None
+
+== Implementation
+* create a new classes called Message and Event in common.
+* Message object is populated with json and validated by a validator.
+* Message object structure for Operation
+```
+message_id : "uuid",
+priority: "", (no priority)
+timestamp: time_now,
+publisher: "node_agent/ceph_integration......" (particular module name)
+"job_id": "",
+"node_id": "",
+"cluster_id": "",
+"flow_id": "",
+"parent_id",
+payload: {
+    "message": ""
+    "exception_traceback": "" (created automatically only for exception messages)
+    "exception_type": "" (created automatically only for exception messages)
+    "exception": "" (Only for exception messages)
+}
+```
+* Message structure for Alert (alert and normal log structure is same, only difference
+between alert and normal log message is priority, Alert always NOTICE)
+```
+message_id : "uuid",
+priority: "notice", (always notice)
+timestamp: time_now,
+publisher: "node_agent/ceph_integration......" (particular module name)
+node_id: "",
+cluster_id: "",
+payload: {
+    message: "" (json of all alert details from collecd, alert handler will
+                 tackcare for alert structure)
+}
+```
+* Message structure for normal log
+```
+message_id : "uuid",
+priority: "",
+timestamp: time_now,
+publisher: "node_agent/ceph_integration......" (particular module name)
+node_id: "",
+cluster_id: ""
+payload: {
+    message: "",
+}
+```
+* Message object is passed to event class and event class will takecare of connecting
+  socket and put the object into socket.
+* Node agent needs to implement a message handler to create, start socket and read a message
+  object from it.
+* Message handler will pass the message object into logger module and the logger
+  figures out where the object needs to be dumped:
+```
+- Every message is dumped to syslog.
+- Messages with a job id are dumped to the operation's update directory.
+- store all the message in etcd.
+    1.global level - all messages
+    2.node level - If node_id is present
+    3.cluster level - If cluster_id is present
+```
+* Message priorities and usages:
+```
+- Debug: Low level information and implementation details that are never pushed to etcd.
+  This includes backtraces.
+- Info: Generic messages. Any messages about flow invocations should be info.
+  Messages originating from storage system or monitoring events may end up here.
+  This is also not stored in etcd.
+- Notice: Hardcoded messages that need to be pushed as notifications.
+  Messages originating from storage system or monitoring events may end up here.
+- Warn: Hardcoded messages that are pushed as UI notifications, but contain
+  warnings instead of just information. Messages originating from storage system
+  or monitoring events may end up here.
+- Error: Errors generated during execution. Probably as a result of exception
+  handling. These may original from within or outside Tendrl.
+- Critical: Operational problems, such as process not responding or hardware
+  failures, critical threshold breaches etc.
+```
+
+== Assignee(s)
+
+Primary assignee: Gowtham Shanmugasundaram
+
+Other contributor(s): Anmolsachan
+
+== Work Items
+
+* https://github.com/Tendrl/specifications/issues/28
+
+== Dependencies
+
+None
+
+== Testing
+
+* Sanity check for flow.
+* Check all log messages are in correct format
+* Check alert is passed to socket successfully
+
+== Documentation impact
+
+None
+
+== References
+
+* https://github.com/Tendrl/specifications/issues/55
+* https://github.com/Tendrl/specifications/blob/master/specs/pluggable_alert_delivery.adoc
+* Tendrl/usmqe-tests#14 for testing
--- a/specs/logging_implementation.adoc
+++ b/specs/logging_implementation.adoc
@@ -1,4 +1,4 @@
-= Generic message mechanism for logs and alerts in Tendrl"
+= Generic message mechanism for logs and alerts in Tendrl
 
 Logs and alerts need to be treated as messages emitted by Tendrl.
 


GowthamShanmugam:
@brainfunked @shtripat please review

GowthamShanmugam:
@shtripat please review this

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
+= Specification for logging implementation
+
+Logs and alerts need to be treated as messages emitted by a component.
```
shtripat: "...emitted by Tedrl modules" may be I would say

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
+= Specification for logging implementation
```
shtripat: "Generic message mechanism for logs and alerts in Tendrl"

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+Logs and alerts need to be treated as messages emitted by a component.
+
+== Problem description
```
shtripat: The problem statement for this case is something like

```
The logs and alerts in Tendrl are handled differently and there is no standard mechanism which makes sure no internal details are exposed as part of alerts/notifications sent to end user.

Ideally logs messages and alerts are both messages which are emitted from an application. Tendrl needs to treat them as one entity and a common handler should decide what to be done with the message which is emitted and received at handler."
```

The below text talks more of implementation how and what you want to do rather :)

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+* For messages that serve as notifications to the administrator (or alerts),
+  exact error message must be hardcoded by the developers. No backtraces or
+  system generated error messages should be sent as notifications to the
+  administrator.
+* Notifications sent to the administrator should provide sufficient context to
+  enable debugging of the problem via investigation into the log files.
+* Notifications and alerts need to be delivered to the node agent alert sockets.
+* Commons needs to implement a logging library that hides all the implementation
+  details and configuration options from any of the components using the library.
+  Components should be able to blindly emit messages to the common logging module
```
shtripat: We should rather call it Message Handler module in Tendrl I feel

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+  sockets etc. depending upon the metadata and configuration.
+* Syslog must be the default logging configuration. Logging to individual
+  configuration files should be thoroughly discouraged.
+
+== Use Cases
+
+Logs and alerts need to be treated as messages.
+
+== Proposed change
+* Create a new class called "Message" in common with attributes name, type,
```
shtripat: I feel we would need info about the module which is raising this message so that module specific logging configuration can be used for logging the messages. I am coming from a point where each module like node_agent, ceph_integration, gluster_integartion etc can have their own log-levels and loggers defined.
Does not look very clear how this scenario will be taken care with this model..

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+* Import the Message class in all the modules and Change log statements to
+  populate the instance of message class and pass the object to function.
+* Expected output format for log and alert in log file:
+```
+Format: "%(asctime)s - %(levelname)s - %(message)s"
+Message: "%(filename)s - %(function)s - %(lineno)s - %(message)s"
+```
+* Expected output format for the alert in node_agent socket:
+```
+{
```
shtripat: This looks very minimal and alerts might need additional fields related to acknowledgment etc. You may refer skyring for detailed object structure may be.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+== Impacted modules
+
+=== Tendrl/common impact
+
+* Create a new class called "Message" in common with attributes name, type,
+  message, priority(empty for log), alert_id ,log_level(empty for alert),
+  function_name, line_number.
+* Using name retrieve the particular logging handler.
+* Create a new function for Message class and process the object values.
```
shtripat: I feel there should be a generic class defined as MessageHandler and extend this into two classes namely LogMessageHandler and AlertHandler. This class would have a handle() function which is specific in nature and takes care of the actual logic.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+* If the type is Alert then create a new uuid and assign the uuid to alert. After
+  that convert the object into json, If the json is created successfully then store
+  json into log file and put the json in node_agent socket also.
+* If the type is Log then process the object and store json in log file based on
+  log level.
+
+=== Tendrl/node_agent impact
+
+* Import the Message class in all the modules and Change log statements to
```
shtripat: I feel we should add the framework in one PR and do the mass change in different PR. Makes life easier and doesnt halt everybody else :)

specs/logging_implementation.adoc
GowthamShanmugam: i will change it

specs/logging_implementation.adoc
GowthamShanmugam: i will change it

specs/logging_implementation.adoc
GowthamShanmugam: ok :)

specs/logging_implementation.adoc
GowthamShanmugam: done

specs/logging_implementation.adoc
GowthamShanmugam: ok

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked told me we have to log alert also in log file. so spiting like  LogMessageHandler and AlertHandler is good?

specs/logging_implementation.adoc
GowthamShanmugam: ok .. i will change as per your suggestion

specs/logging_implementation.adoc
GowthamShanmugam: ack required for events right?

specs/logging_implementation.adoc
brainfunked: :+1:

specs/logging_implementation.adoc
brainfunked: '..by Tendrl'?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+Logs and alerts need to be treated as messages emitted by a component.
+
+== Problem description
+
+* Logs and alerts need to be treated as messages emitted by a component.
+* A schema needs to be defined for emitted messages. This could be a json schema.
+  The schema needs to include metadata that allows the exact point of origin of
+  the message to be known.
+* Message priorities, types etc. need to be defined at the time of emission.
```
brainfunked: '..at the point of origin'.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+* Notifications sent to the administrator should provide sufficient context to
+  enable debugging of the problem via investigation into the log files.
+* Notifications and alerts need to be delivered to the node agent alert sockets.
+* Commons needs to implement a logging library that hides all the implementation
+  details and configuration options from any of the components using the library.
+  Components should be able to blindly emit messages to the common logging module
+  and it should be able to deliver them to either syslog, log files, alert
+  sockets etc. depending upon the metadata and configuration.
+* Syslog must be the default logging configuration. Logging to individual
+  configuration files should be thoroughly discouraged.
```
brainfunked: '..discouraged by default.'

specs/logging_implementation.adoc
brainfunked: Additionally, the updates to in-flight operations, https://github.com/Tendrl/specifications/issues/55, need to be logged under INFO. The updates themselves should just be messages emitted that have the same structure and appropriate metadata. This metadata includes operation uuids. This implies that the details of the jobs being created also need to be logged. This is where tracing and debugging is addressed. The updates being logged provide an id. This id pertains to a specific job. This job being logged enables end-to-end tracing of the job and it's updates across the deployment. This is very important, since it allows an administrator to filter the logs by id and get all the required details.

brainfunked (changes requested):

specs/logging_implementation.adoc
GowthamShanmugam: yes i feel same .. we cant split

specs/logging_implementation.adoc
GowthamShanmugam: done

specs/logging_implementation.adoc
GowthamShanmugam: done

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
+= Specification for logging implementation
+
+Logs and alerts need to be treated as messages emitted by a component.
+
+== Problem description
+
+* Logs and alerts need to be treated as messages emitted by a component.
```
t0ffel: Apologies in advance as I am a bit out of context on Tendrl, but what is the difference between logs and alerts within the scope of this specification?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+* Syslog must be the default logging configuration. Logging to individual
+  configuration files should be thoroughly discouraged.
+
+== Use Cases
+
+Logs and alerts need to be treated as messages.
+
+== Proposed change
+* Create a new class called "Message" in common with attributes name, type,
+  message, priority(empty for log), alert_id ,log_level(empty for alert),
```
t0ffel: Several questions:
1. What about `timestamp`?
2. Is there a place for extra metadata that might accompany the message?
3. Do you plan to have possible values/expected format for these fields? i.e.
    * does the `log_level` have 1-1 mapping to syslog's `severity`?
    * What is the format of the `message`? Can it be a multiline backtrace?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+```
+1. If the type is Alert then create a new uuid and assign the uuid to alert. After
+   that convert the object into json, If the json is created successfully then store
+   json into log file as warning and put the json in node_agent socket also.
+2. If the type is Log then process the object and store json in log file based on
+   log level.
+```
+* Import the Message class in all the modules and Change log statements to
+  populate the instance of message class and pass the object to function.
+* Expected output format for log and alert in log file:
```
t0ffel: what about sending logs to syslog? What will the logs/alerts look like in syslog/journal? Will you be using any special systemd journal fields?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,174 @@
...
+```
+Format: "%(asctime)s - %(levelname)s - %(message)s"
+Message: "%(filename)s - %(function)s - %(lineno)s - %(message)s"
```
t0ffel: I'm a bit confused. In https://github.com/Tendrl/specifications/pull/94/files#diff-5a251eddb4570c9ab1b30f91c2d17180R54 it says that json will be stored. 
Also what exactly do you mean by separate lines for `Format` and `Message`?

t0ffel (commented):
Hi, I'm with Common Logging. I was asked to take a look at this PR.
I'm not very familiar with tendrl, so please forgive me if my questions occasionally sound stupid.

specs/logging_implementation.adoc
GowthamShanmugam: Log nothing but a information, erros. etc. It helps  an end_user for the debugging the problem. Alert also a message but it is immediately forward to end_user. Alerts can be debug with the help of log message.

specs/logging_implementation.adoc
GowthamShanmugam: 1.ya i have toadd timestamp also,
2.I have declared class called Message for holding the message and other details. And alert and log can be handled from that object.
3.
    * yes log_level have 1-1 mapping to syslog's : log level also came from object, with the help of log level i am calling syslog function.
@brainfunked your comments pleases

specs/logging_implementation.adoc
GowthamShanmugam: What is my idea is Create class called Message, This message class have all the attributes like 
               "name" (log_handler_name) (each module have different logging conf, so we are going to log                      message in common place, so we need log_handler for that corresponding module).
                "type" (alert or log) (we log alert also but alert have one more step, alert is passed to enduser immediately ).
                 "log level" - (it tells what is the log_level like info, error.... for log the message in syslog, alert is default warning).
                 "timestamp"
                 "severity" (for log it is empty)
                 "aler_id"
                 "tag" (for alert)
                 "message" - message is nothing but text what they expected to log or alert. Developer will hard code the message. Why i gave message format means 
if you give log format Format: "%(asctime)s - % (filename) - %(function) - %(linenumber) %(levelname)s - %(message)s" then log message logged with filename, functionname, loneno where the loglevel function called.

Here we are collecting all object in handler function (in message class) and logging from that function itself. So log message will take handler function's name, filename, lineno in log file. To avoid this the handler will collect the filename, function name, lineno automatically where the object is came from and it will append these details with message and then log that message in log file. So that appending format is    Message: "%(filename)s - %(function)s - %(lineno)s - %(message)s".

we are using message object to collect all info, no need any json, i have to modify that line

GowthamShanmugam: format is nothing but we give logging format when we create log handler,
message format nothing but appending extra info with message automatically

specs/logging_implementation.adoc
GowthamShanmugam: no we are not using any special journal, 
using loglevel we are calling exact syslog function to log the message

alert is always warning. For log Based n loglevel it is decided (info, error....)

GowthamShanmugam:
@shtripat @brainfunked @nnDarshan @anmolbabu please review this

GowthamShanmugam:
@brainfunked @shtripat @anmolbabu @nnDarshan

GowthamShanmugam:
@brainfunked @shtripat @nnDarshan please review this

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,201 @@
+= Generic message mechanism for logs and alerts in Tendrl"
+
+Logs and alerts need to be treated as messages emitted by Tendrl.
+
+== Problem description
+* The logs, alerts and flight_operations_updates in Tendrl are handled differently
```
shtripat: `in flight operations updates`

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,201 @@
...
+  timestamp. (other attributes are filename, function_name, line_number, timestamp.
+  These details are appended with message automatically).
+```
+Message class object structure
+     log_handler_name
+     message_type(log/alert/flight_operation)
+     log_level(info/error..)
+     message
+     message_trace_back_flag(True/False)
+     message_uuid (optional for log)
```
shtripat: It should not optional for logs may be we should use current job id as message id so that tracing of the error becomes easier while debugging.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,201 @@
...
+    2. For flight_oper - message_handler(log_handler_name, message_type, log_level, message,
+       message_trace_back_flag, message_uuid(job_id))
+    3. For alert - message_handler(log_handler_name, message_type, log_level, message,
+       message_trace_back_flag, message-uuid(alert_id), severity, tag)
+```
+* Create a new function called message_handler in Message class and process the
+  object values.
+* If the message type is Alert then store the message in log file with uuid (log level
+  for alert is always warning). and put the object in node_agent socket.
+* If the message type is flight_operation then log the message in file with message
```
shtripat: as INFO

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,201 @@
...
+  uuid (uuid is used to trace the filght operation).
+* If the type is Log then process then store message in log file.
+
+=== Tendrl/node_agent impact
+
+* Create socket to receive alert.
+* Import message class from common and change all log statement (it is handled in another PR).
```
shtripat: Put the link to other PR here

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,201 @@
...
+  2. If the message type is flight_operation then log the message in file with message
+     uuid (uuid is used to trace the filght operation).
+  3. If the type is Log then process then store message in log file..
+  ```
+  * Expected output format for log, alert, flight_operation in log file:
+  ```
+  Message: ""%(asctime)s - %(filename)s - %(function)s - %(lineno)s - %(levelname)s
+  - %(message)s"
+
+  (No need to specify any log format in configuration file. We can remove the format
```
shtripat: Need to POC this I feel

specs/logging_implementation.adoc
GowthamShanmugam: done

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat  yes actually my logic also same why i mentioned here as optional means, for flight operation i am using job id as message id, for alerting i am using alert id as message id but what about normal exception logs. It does not required any id. What i try to mentioned here is for normal log messages it is optional

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat ok  i will change it

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat For this specification is any required?

shtripat: All I meant is instead if saying handled in another PR put the reference of that PR

specs/logging_implementation.adoc
shtripat: May be make the sentences more clearer in that case

GowthamShanmugam: @shtripat ok

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat ok i will do it

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat done

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+== Problem description
+* The logs, alerts and flight operations updates in Tendrl are handled differently
+  and there is no standard mechanism which makes sure no internal details are
+  exposed as part of alerts/notifications sent to end user.
+* Ideally logs messages, flight operation updates and alerts are messages which
+  are emitted from an application. Tendrl needs to treat them as one entity and
+  a common handler should decide what to be done with the message which is emitted
+  and received at handler.
+== Use Cases
```
anmolbabu: You probably need to leave a line above

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,202 @@
...
+  handler_name, message_type, message_uuid, log_level, message_tag, message, severity,
+  timestamp. (other attributes are filename, function_name, line_number. These are
+  populated by message handler automatically).
+  (Logging will take line number and function name from where it is called, Here
+  the problem is we are logging the message from common so it takes a function name
+  and line number from common only. So it makes debugging complex. To avoid this
+  we have to find the function name, line number, file name and append this data
+  with message manually).
+  ```
+  from inspect import getframeinfo, stack
```
anmolbabu: This is not looking properly in html generated from adoc use asciidoctor <file_name>.adoc after installing asciidoctor. The command generates a html file in current directory and that you can open in any browser and check it. You probably need to leave a line before ```` or not sure if ```` is a valid symbol in adoc. Please check it.

specs/logging_implementation.adoc
GowthamShanmugam: @anmolbabu done

specs/logging_implementation.adoc
GowthamShanmugam: @anmolbabu this symbol should start from first column without any space, then it will work

shtripat: @GowthamShanmugam better follow a line gap and start from first column of line always.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+* Import message class from common and change all log statement (it will be in another PR).
+
+== Security impact
+
+None
+
+== Notifications/Monitoring impact
+
+None
```
anmolbabu: We should probably say:

The collectd plugin to handle collectd detected threshold breaches should be customized to call the utility function described in this adoc instead of .directly posting the alert to socket

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,202 @@
...
+* If the type is Log then store the message in log file based on log level.
+
+=== Tendrl/node_agent impact
+
+* Create socket to receive alert.
+* Import message class from common and change all log statement (it will be in another PR).
+
+=== Tendrl/gluster_integration impact
+
+* Import message class from common and change all log statement (it will be in another PR).
```
anmolbabu: We should also be adding this:

gluster-integration needs to invoke the utility described in this PR for every alert detected by it.

anmolbabu: Similarly for ceph-integration also

specs/logging_implementation.adoc
anmolbabu: @GowthamShanmugam please check https://github.com/GowthamShanmugam/specifications/blob/4bdc50f125b05162fcf36117266b2fcba21ae745/specs/logging_implementation.adoc
for seeing how the code part looks currently. It looks mis-aligned and not as intended.

specs/logging_implementation.adoc
GowthamShanmugam: ok

specs/logging_implementation.adoc
GowthamShanmugam: i will add this

specs/logging_implementation.adoc
GowthamShanmugam: i will correct it

GowthamShanmugam:
@brainfunked @shtripat @nthomas-redhat @nnDarshan @anmolbabu please review this

mkudlej:
Please add https://github.com/Tendrl/usmqe-tests/issues/14 as reference for testing alerts.

GowthamShanmugam:
@mkudlej done

GowthamShanmugam:
@anmolbabu @shtripat @brainfunked @nthomas-redhat please review

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+* Message priorities, types etc. need to be defined at the time of emission.
+* The integration/bridge applications must pass the message as per schema.
+* For messages that serve as notifications to the administrator (or alerts),
+  exact error message must be hardcoded by the developers. No backtraces or
+  system generated error messages should be sent as notifications to the
+  administrator.
+* Notifications sent to the administrator should provide sufficient context to
+  enable debugging of the problem via investigation into the log files.
+* Notifications and alerts need to be delivered to the node agent alert sockets.
+* Commons needs to implement a logging library that hides all the implementation
```
shtripat: commons module??

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+  configuration should be used to log the messages, So we have to pass logging
+  handler name of the perticular module to Message Handler. Message handler will
+
+* Create a new function for Message class and process the object values.
+```
+  1. If the message type is Alert then store the message in log file with uuid (log level
+     for alert is always warning). and put the object in node_agent socket.
+  2. If the message type is flight operation then log the message in file with message
+     uuid (uuid is used to trace the filght operation).
+  3. If the type is Log then process then store message in log file..
```
shtripat: grammar?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+     "alert_id"       : "string",
+     "node_id"        : "string",
+     "resource"       : "string",
+     "current-value"  : "string",
+     "significance"   : "string"
+     "severity"       : "string",
+     "type"           : "string",
+     "tags"           : "string",
+     "ackedby"        : "string",
+     "acked"          : "boolean"
```
shtripat: Should we have ack_comment as well here?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+         caller = getframeinfo(stack()[1][0])
+         self.filename = caller.filename
+         self.function_name = caller.function
+         self.line_number = caller.lineno
+         self.timestamp = time.time()
+         # logic to handle alert, log and flight operation
+
+import uuid
+
+class alert(Message):
```
shtripat: Class name starts with caps

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat *common

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat i will change it

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat i will change it

specs/logging_implementation.adoc
GowthamShanmugam: @shtripat ack_comment? why this is required , @anmolbabu your suggestion please

shtripat: While acking a notification user might provide some comment and that we can store.

specs/logging_implementation.adoc
shtripat: I mean you should say "common module"

specs/logging_implementation.adoc
GowthamShanmugam: @anmolbabu your suggestion please

anmolbabu: @GowthamShanmugam @shtripat comment looks like is required but I missed it somehow in my spec that actually talks about alert structure. But its my opinion that, instead of having changes to the alert structure and details around it be effected in 2 places this spec and also https://github.com/Tendrl/specifications/blob/master/specs/pluggable_alert_delivery.adoc. So I feel we can open a bug to capture this point in pluggable_alert_delivery.adoc and just refer that spec in this spec for details about alerts. @shtripat what do you suggest

specs/logging_implementation.adoc
anmolbabu: Please refer my comment @ line 208

GowthamShanmugam: @anmolbabu ok

specs/logging_implementation.adoc
GowthamShanmugam: LGTM

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+  in Tendrl and it should be able to deliver them to either syslog, log files,
+  alert sockets etc. depending upon the metadata and configuration.
+* Syslog must be the default logging configuration. Logging to individual
+  configuration files should be thoroughly discouraged.
+
+== Alternatives
+
+* None
+
+== Data model impact
```
nnDarshan, line 40: should you not add a model for alerts in the definition file as it is stored in etcd ?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+None
+
+== Developer impact
+
+None
+
+== Implementation
+
+* Create a new class called "Message" in common for message handling with attributes
+  message_type, filename, function_name, line_number, timestamp.
```
nnDarshan: add more details about timestamp, like it has to be in UTC format etc

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+  - message_type
+  - filename
+  - function_name
+  - line_number
+  - timestamp
+```
+* Create a new function called message_handler in Message class and process the
+  object values.
+* If the message type is Alert then store the message in log file with uuid (log level
+  for alert is always warning). and put the object in node_agent socket.
```
nnDarshan: why is it always warning ? it can be other levels as well right ?

nnDarshan: log level can be different right ? why always warning ?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+== Data model impact
+
+None
+
+== Impacted modules
+
+=== Tendrl/common impact
+* Create a new class called "Message" in common for message handling with attributes
+  message_type, filename, function_name, line_number, timestamp.
```
nnDarshan: add details about time-stamp in which format ? like UTC

specs/logging_implementation.adoc
nnDarshan, line 40: you need to add a entry for model of alerts in the definition file right as it is stored in etcd ?

GowthamShanmugam: @nnDarshan No i am going to put alert details in node socket only, node socket will tack care for storing in etcd

GowthamShanmugam: @nnDarshan I am not going to store alert details in etcd, my code will put that in node socket only

specs/logging_implementation.adoc
GowthamShanmugam: @nnDarshan i will add it

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+  (No need to specify any log format in configuration file. We can remove the format
+  line from configuration files)
+```
+* Expected output format for the alert in node_agent socket:
+```
+  {
+     "timestamp"      : "time"
+     "alert_id"       : "string",
+     "node_id"        : "string",
```
anmolsachan: Should we have cluster_id also ? Also if we add cluster_id, will there be any need of node_id ? @shtripat your suggestions please.

specs/logging_implementation.adoc
GowthamShanmugam: @nnDarshan Levels only for normal logs, Alerts is a warning right? how alert have other log levels

GowthamShanmugam: only normal logs have different levels, alerts always in warning only

specs/logging_implementation.adoc
GowthamShanmugam: @nnDarshan i will add it

specs/logging_implementation.adoc
GowthamShanmugam: i think cluster is is required, node is required or not ? @shtripat

shtripat: What happen if the alert belongs to a node which is not yet participating in cluster. I feel node_id is fine and once node participates in cluster, using node_id we can always get the cluster to which it belongs to.

GowthamShanmugam: @anmolsachan cluster id is comes with tags

specs/logging_implementation.adoc
GowthamShanmugam: i have discussed with @shtripat ,  @anmolbabu and @nnDarshan. Alert needs different log levels.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,284 @@
...
+  handler_name, message_type, message, log_level, timestamp(format: ISO 8601).
+* Create a new class called "Log" in common and inherit it from Message class.
+* Create a new class called "Alert" in common with attributes alert_id, node_id,
+  resource, current_value, tags, type, severity, significance, ackedby, acked.
+  Inherit alert class from Message class.
+```
+If it is alert then check it is clear alert or not.
+ - Pass the alert details to the function which is declared in socket handler module to
+   check it is clear alert or not. If yes then log the alert as INFO otherwise log the
+   alert based on severity.
```
GowthamShanmugam: I need suggestion for this approach @shtripat @anmolbabu @nnDarshan please share your suggestion

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,284 @@
...
+ - loglevel for alert is based on alert severity
+   - If alert severity is Warning then loglevel is WARNING.
+   - If alert severity is Critical then loglevel is CRITICAL.
+   - If alert severity is Ok then loglevel is INFO.
```
GowthamShanmugam: I need suggestion for this approach @shtripat @anmolbabu @nnDarshan please share your suggestion

shtripat: LGTM

anmolbabu:
@Tendrl/tendrl-core @GowthamShanmugam 

The alert life-cycle involves multiple components as below:
* The alerting module: 
  ** Handles validation of alert.
  ** Processes the alert to find if there was a previously detected similar alert from same source to detect if its a duplicate alert or a clearing alert for the previously occurred alert, etc... and probably a few more will get added as their necessity arises.
  ** Notifying the alert to configured destinations.

* The node-agent:
 ** It takes the only responsibility of performing an initial primitive validation of alert and transporting any validated alert on its socket to etcd.

* The bridge/collectd:
  ** Generate an alert for any status change or utilization threshold breach accordingly and put it on alert socket.

From above, it can be observed that the only place where an alert is in its fully processed/polished form is only in alerting module and everywhere else, its a raw alert that does not have tendrl-specific intelligence processed form of the alert. So my opinion is it would be better to invoke the MessageHandler intended to be developed in accordance with this spec, receive the alert from a place where the alert is actually fully processed(i.e, alerting module) rather than from the source of the alert(where the alert is fairly raw i.e, not processed and validated by tendrl intelligence). 

Please let me know what @Tendrl/tendrl-core feels about this...

specs/logging_implementation.adoc
brainfunked: We also need DEBUG messages for details we'll want to log from the code, that are important enough to allow an administrator to debug or trace a problem, but aren't critical or useful enough as notifications.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,284 @@
...
+```
+* Create a new class called "InFlightOperation" in common with attributes
```
brainfunked: How about just Operation? It doesn't need to be in flight. Anything that relates to an operation should be tracked using the operation id that's trickled down with each call down the stack.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,284 @@
...
+Flight operation class object structure
+  - job_id
+  - operation_summary
+
+Log class object structure
+  - It contains no atrribute, it is based on super class(message). For more info
+    see the sample implementation at the end.
+
+Message class object structure
+  - log_handler_name
```
brainfunked: What is log_handler_name?

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked i will change that as Operation

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked we want to log the massage from common, So log message is logged using modules logging configuration. (e.g) if ceph_integration call massage handler to log some message then log message should goes to ceph_integration log file. so these logging handler is passed from each module with message to log the message in corresponding log file.

GowthamShanmugam:
@brainfunked please give your suggestion for anmol 's comment

specs/logging_implementation.adoc
brainfunked: The alert shouldn't prescribe who should `handle' the alert. It should only specify where the alert originated. The message handler base class should be able to route the alert to the appropriate class based on the metadata, and that's where the alert gets handled.

GowthamShanmugam:
@brainfunked @shtripat @nthomas-redhat @r0h4n @anmolbabu please review

GowthamShanmugam:
@brainfunked i have modified this spec as per your suggestion, please review

GowthamShanmugam:
@shtripat @nthomas-redhat @r0h4n please review

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+```
+1. messages with inorder_id
+    {job_id}/{in_order_id}
+```
+
+== Impacted modules
+
+=== Tendrl/common impact
+* Create a new class called "Message" in common for message handling with attributes
+  handler_name, message_type, message, log_level, timestamp(format: ISO 8601).
```
brainfunked: Just copy the format from [Nova](http://docs.openstack.org/developer/nova/notifications.html), with some additions:
`
{
    "priority": <string, selected from a predefined list by the emitter>,
    "event_type": <string, defined by the emitter>,
    "timestamp": <string, the isotime of when the notification emitted>,
    "publisher_id": <string, defined by the emitter>,
    "node_id": <uuid, defined by the emitter>,
    "message_id": <uuid, generated by the Message object that's instantiated for every single message>,
    "payload": <json serialized dict, defined by the emitter>
        {
            "job_id": <uuid, generated by the API>,
            "parent_job_id": <uuid, generated by the API>,
            "flow_id": <string, fully qualified, namespaced flow name from the definition files, supplied by the API>
        }
}
`

Drop the handler_name first and foremost. The application generating the messages doesn't know who's going to handle the message.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+* Create a new class called "Log" in common and inherit it from Message class.
```
brainfunked: Don't need a Log class. Everything is a log message.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+* Create a new class called "Alert" in common with attributes alert_id, node_id,
```
brainfunked: Don't need an Alert class either. The alerting application should pick the messages up based on the priority.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+  resource, current_value, tags, type, severity, significance, ackedby, acked.
```
brainfunked: Let the alerting module add the metadata as required for handling. This specification is for generating the logs only. The metadata added by the alerting module would include things like acknowledgements, delivery endpoints etc. But that's outside the scope of this specification.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+  Inherit alert class from Message class.
+* Create a new class called "Operation" in common with attributes
```
brainfunked: Drop the Operation class as well. Here's what's actually required:

The `Tendrl_context` object needs to contain the following information:
* job_id
* parent_job_id
* flow_id

This context needs to be passed to each of the atoms and all the way down the call stack. Every method invoked as part of an operation needs to have this context. The logger can then be used to log this information, along with additional metadata, as described in a comment earlier. When a message is logged, the Message class should be instantiated. Upon instantiation, it should generate a UUID for each message. The object should then not be reused. Every message must be a separate object.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+Log class object structure
+  - log_handler_name
+  - log_level
+  - log_message
+
+Message class object structure
+  - message_type
+  - timestamp (format : ISO 8601)
+```
+* Create a new function called message_handler in Message class and process the
```
brainfunked: No. Message class is just for the message itself. The logger module should have a log method which takes the Message object as a parameter.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+  object values.
+* If the message type is Alert then put the object in node_agent socket, logging
```
brainfunked: Again, alert handler comes into the picture AFTER the message has been dumped in etcd.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+  for alert is handled from alert handler.
+* If the message type is operation then log the message based on log level with
```
brainfunked: This'll now change since there's a single Message class. If the message metadata contains a job id, it's an update for an operation, dump it in the operation's sequential directory.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+  message job id (job id is used to trace the operation) and store the object in etcd.
+* If the type is Log then store the message in log file based on log level.
+
+=== Tendrl/node_agent impact
```
brainfunked: Node agent needs to implement a message handler class, which takes a JSON formatted message on it's socket and turns it into a proper Message object. The Message object is then passed to the logger. The logger figures out where the object needs to be dumped:

* Every message is dumped to syslog.
* Messages with a job id are dumped to the operation's update directory.
* Messages with priorities info, notice, warn, error and critical are pushed to an `events` directory in etcd. This directory will replace the `alerts` directory currently used by the alerting module. The alerting module should figure out, based on the priority, what needs to be done with an event.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,273 @@
...
+== Data model impact
+
+* Store the operation details in etcd:
+```
+1. messages with inorder_id
+    {job_id}/{in_order_id}
+```
+
+== Impacted modules
+
```
brainfunked: We'll have the following priorities:

* Debug: Low level information and implementation details that are never pushed to etcd. This includes backtraces.
* Info: Generic messages. Any messages about flow invocations should be info. Messages originating from storage system or monitoring events may end up here.
* Notice: Hardcoded messages that need to be pushed as notifications. Messages originating from storage system or monitoring events may end up here.
* Warn: Hardcoded messages that are pushed as UI notifications, but contain warnings instead of just information. Messages originating from storage system or monitoring events may end up here.
* Error: Errors generated during execution. Probably as a result of exception handling. These may original from within or outside Tendrl.
* Critical: Operational problems, such as process not responding or hardware failures, critical threshold breaches etc.

brainfunked (changes requested):
@r0h4n @anmolbabu Please review.

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked operations also have priorities, current job update have priority also

nnDarshan: @brainfunked Any specific reason for using sockets in node_agent?  Cant node_agent message handler class expose a method that can be called to push the message ?

brainfunked: @nnDarshan It's because other processes would want to be able to push their messages through the node agent. Anyway, @anmolbabu has already worked on the alerts socket, the same can be resued.

brainfunked: @GowthamShanmugam I'm not sure what you mean. Can you provide examples?

GowthamShanmugam: @brainfunked in current job steps update @r0h4n is storing message with key wise
_jobs/tendrl.gluster_integration.flows.stop_volume.StopVolume_f48c70f6-e2ed-4620-9256-ddf020a6b64d/warn"

_jobs/tendrl.gluster_integration.flows.stop_volume.StopVolume_f48c70f6-e2ed-4620-9256-ddf020a6b64d/error

_jobs/tendrl.gluster_integration.flows.stop_volume.StopVolume_f48c70f6-e2ed-4620-9256-ddf020a6b64d/all
_jobs/tendrl.gluster_integration.flows.stop_volume.StopVolume_f48c70f6-e2ed-4620-9256-ddf020a6b64d/info

GowthamShanmugam: @brainfunked 

"key": "/clusters/2a4b84e0-17b3-4543-af9f-e42000c52bfc/_jobs/tendrl.gluster_integration.flows.stop_volume.StopVolume_f48c70f6-e2ed-4620-9256-ddf020a6b64d/info",
"value": "[u'Processing pre-runs for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', 'Start pre-run : tendrl.gluster_integration.objects.Volume.atoms.volume_exists.VolumeExists', u'Finished pre-run: tendrl.gluster_integration.objects.Volume.atoms.volume_exists.VolumeExists for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', 'Start pre-run : tendrl.gluster_integration.objects.Volume.atoms.volume_started.VolumeStarted', u'Finished pre-run: tendrl.gluster_integration.objects.Volume.atoms.volume_started.VolumeStarted for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', u'Processing atoms for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', 'Start atom : tendrl.gluster_integration.objects.Volume.atoms.stop.Stop', u'Finished atom tendrl.gluster_integration.objects.Volume.atoms.stop.Stop for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', u'Processing post-runs for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume', 'Start post-run : tendrl.gluster_integration.objects.Volume.atoms.volume_stopped.VolumeStopped', u'Finished post-run: tendrl.gluster_integration.objects.Volume.atoms.volume_stopped.VolumeStopped for flow: tendrl.gluster_integration.flows.stop_volume.StopVolume']",

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked This job_id and parent_id and flow_id is for currently running job or i have to store all the job in tendrl context

specs/logging_implementation.adoc
cybernth: As discussed, let us make 'Notice' as the default log level so that lower level severities like info, debug are not logged by default.

GowthamShanmugam:
@brainfunked @nthomas-redhat @shtripat @nnDarshan @anmolbabu @anivargi @r0h4n please review this spec , i have updated as per discussion.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+      timestamp: time_now,
+      publisher: "node_agent/ceph_integration......" (particular module name)
+      payload: {
+         "node_id": "",
+         "cluster_id": "",
+         "message": """
+      }
+   }
+
+    (operational messages stored in operational directory only)
```
brainfunked: How? Where? Document the details please.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+```
+
+== Impacted modules
+
+=== Tendrl/common impact
+* Create a logger module in common to handle the message handler object.
+* Logging module will takecare of log the messages in log file.
+* Create a new function called emit in logger module to process message object.
```
brainfunked: emit = generate, handle = process. What precisely do you mean for the emit method to do?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+```
+ - If the message have "job_id" then it is considered as operation, store message
+   in operation directory in etcd with sequential key and log the message with job
+   id in log file. (Always log the operation messages as info).
+ - If the message have priority then logger will store the message in etcd (except info and debug).
```
brainfunked: Every message must have a priority. If not, the entire message needs to logged as a warning - a warning message, containing another message that isn't well formed.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+          In global level : /messages/{message_id}
+          In node level   : /nodes/{node_id}/messages/{message_id}
```
brainfunked: If the message has a node_id, always copy it under the node's namespace.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+=== Tendrl/common impact
+* Create a logger module in common to handle the message handler object.
+* Logging module will takecare of log the messages in log file.
+* Create a new function called emit in logger module to process message object.
+```
+ - If the message have "job_id" then it is considered as operation, store message
+   in operation directory in etcd with sequential key and log the message with job
+   id in log file. (Always log the operation messages as info).
+ - If the message have priority then logger will store the message in etcd (except info and debug).
+          In global level : /messages/{message_id}
```
brainfunked: Always log every message globally.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+          In node level   : /nodes/{node_id}/messages/{message_id}
+          In cluster level: /clusters/{cluster_id}/messages/{message_id}
```
brainfunked: If the message has a cluster_id, always copy it under the cluster's namespace.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+ - Store the all message in log file based on log level.
```
brainfunked: Push all the log messages via syslog and don't bother with log files.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+```
+
+=== Tendrl/node_agent impact
+* Add job id, flow id, parent id in tendrl context.
+* Create a new class for message_handler to hold the hold the values.
```
brainfunked: What does this mean?

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+* Now the alert handler will now be used generically for all three. (alert, log and operation).
```
brainfunked: Message handler. But you've still not specified exactly what it's supposed to do and where it gets invoked.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,262 @@
...
+* The alert handler will changed to message handler to handle all messages.
+* Message handler should read the json from node agent socket and populate the message
+  handler object
+```
+structure of message handler
```
brainfunked: Structure of the Message object!

brainfunked: Handler should just read from the socket, read the json into a Message object and run validations. The Message object itself should have validations to ensure that it's well formed. So the handler essentially does Message.new(json).validate or some such. At this point, you have a Message object that's either valid or invalid. The invalid one gets logged as a warning by creating another Message object that contains the entire original Message object as nested payload. The valid one gets passed to the logger, which figures out where it needs to be copied to: multiple locations in etcd and always via syslog.

brainfunked (changes requested):

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked have mentioned in data model section where operation message will store

GowthamShanmugam:
@brainfunked @nthomas-redhat @shtripat @anmolbabu @nnDarshan @anivargi @r0h4n please review this

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,311 @@
...
+          "job_id": "",
+          "node_id": "",
+          "cluster_id": "",
+          "flow_id": "",
+          "parent_id",
+          "message": ""
+       }
+    }
+
+2. All Messages are stored in etcd
```
brainfunked: Correction: all messages are sent to syslog, only notice and above are stored in etcd.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,311 @@
...
+    "payload": <json serialized dict, defined by the sender>
+    (payload will change for log and operation other details are common)
+```
+* The Message object itself should have validations to ensure that it's well formed.
+  The validator should check the object contain all the necessary information are populated
+  correctly. If the object is invalid then it create a new message object and store the
+  original object under payload in newly created object and set the priority as warning
+  and return the newly created object. If the object is valid then it return same object.
+* When the message object gets populated then it is passed to event class.
+* Create a new class called Event. Event class will tackcare of pass the message object
```
brainfunked: Event class should emit a Message object.

specs/logging_implementation.adoc
GowthamShanmugam: done

specs/logging_implementation.adoc
GowthamShanmugam: done

brainfunked (approved):

GowthamShanmugam:
@TimothyAsir @brainfunked @anmolbabu @nthomas-redhat @shtripat At the time of logging a message, if the socket fails then log message never reach to logging framework. In this case we have to intimate the end user like socket is down and log messages are not logged successfully. Other wise the log message will lost. The problem here is without socket we can log the message anywhere, we need some mechanism to handle this case.

shtripat:
@GowthamShanmugam I am not sure how to tackle, but as a preferable option I would like to log these kind of failures in some common tendrl files say `/var/log/tendrl/tendrl.log`. Just a wierd thought...

cybernth:
> @TimothyAsir @brainfunked @anmolbabu @nthomas-redhat @shtripat At the time of logging a message, if the socket fails then log message never reach to logging framework. In this case we have to intimate the end user like socket is down and log messages are not logged successfully. Other wise the log message will lost. The problem here is without socket we can log the message anywhere, we need some mechanism to handle this case.

@GowthamShanmugam 
In case of logger failure :
log errors to stderr and messages to stdout , this should ultimately go into systemd journals.
we should be able to alert to etcd directly via the node agent in case there's a problem with the socket .
Errors should be logged to stderr always regardless of logger failed or not

GowthamShanmugam:
@nthomas-redhat i will change as per your suggestion

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,308 @@
+= Generic message mechanism for logs and alerts in Tendrl"
```
brainfunked: Remove the " at the end.

specs/logging_implementation.adoc
```diff
@@ -0,0 +1,308 @@
...
+
+== Alternatives
+
+* None
+
+== Data model impact
+
+* Store the operation details in etcd:
+```
+1. Operation messages with inorder_id
```
brainfunked, line 44: The etcd message directory locations are missing. Add them.

For job specific updates, the following structure is expected:
* `/queue/<job_id>` where `job_id` is the UUID of the job with no prefixes or suffixes.
* `/queue/<job_id>/parameters` containing all the parameters for the job.
* `/queue/<job_id>/messages` containing the sequentially generated keys, each containing a single message pertaining to the job.

In addition to these, there needs to be a specific file that updates with the job's overall status and marks the job as finished, either successfully or with errors, with the error details etc. @anivargi your inputs are needed here. The concern is to avoid multiple queries per poll for a job, if possible.

brainfunked (changes requested):

specs/logging_implementation.adoc
GowthamShanmugam: @brainfunked job_key is not available for flows, i had discuss with @shtripat and @nnDarshan they said when the job is received by component update parameter with job_key.   
Because no need to change signature of any function for call job_key separately.

GowthamShanmugam: @brainfunked 
My understanding is  all the parameter of jobs are stored under a non-sequential key  i.e
/queue/<job_id>/parameters, this is done by api during job submission. i have to this with @anivargi  help.

Message handler module should place the job updates  under /queue/<job_id>/messages/<seq-key> for particular job.

And /queue/<job_id>/finished contains job status like job finished with error, errors, etc... 

is this correct?

anivargi: @GowthamShanmugam 
Here is how it needs to be
`/queue/<job_id>` will be a directory
`/queue/<job_id>/status` is the status of the current job i.e new, processing, finished, error
`/queue/<job_id>/errors` is the errors caused in the job if the status is error.
`/queue/<job_id>/parameters` is the parameters of the job which is queued.
`/queue/<job_id>/messages` will be a directory containing all the messages.
`/queue/<job_id>/messages/sequence` will be the sequence number and value will be the message string

Other possible attributes under `/queue/<job_id>/..` could be integration_id, run, flow, type, node_ids.

Hope its clear now.
//@brainfunked

GowthamShanmugam:
@brainfunked @r0h4n @shtripat Spec is updated

specs/logging_implementation.adoc
r0h4n: @anivargi where would the "run" , "node_ids" etc items go as per this new structure?

anivargi: @r0h4n the run would come in as `/queue/<job_id>/run` and node_ids will come in as `/queue/<job_id>/node_ids`. Note that node_ids would be a comma separated value since we cannot represent an array in etcd value like we do in json today.

anivargi:
@GowthamShanmugam slight changes are going to be have to made as follows:

`/queue/:job_id` will be a directory
`/queue/:job_id/payload` will be the json dump of the job with the same structure we have today
`/queue/:job_id/status` will be the global status which needs to be updated as per the currentstatus of the job, default will be 'new'
`/queue/:job_id/errors` will be a message if the status is `error` 
`/queue/:job_id/messages` will be the log messages related to the job.

The request_id is not need anymore for the API to know about the job messages.
CC: @r0h4n @brainfunked

anivargi:
@brainfunked @GowthamShanmugam I would also like to recommend if we could put the messages under something like `/messages/jobs/:job_id`, would that be a option?

anivargi:
@GowthamShanmugam We need to move the messages per job under `/messages/jobs`

`/messages/jobs/:job_id` will be a directory
`/messages/jobs/:job_id/:sequence` will have value of the message

Other non job related messages will reside under 
`/messages/events`

CC: @brainfunked

GowthamShanmugam:
@brainfunked @anivargi @r0h4n as per above discussion i have modified job structure  https://github.com/Tendrl/commons/pull/175/files

r0h4n:
@anivargi I think we have missed the "parent_id" field in this job structure

GowthamShanmugam:
@r0h4n @anivargi as per current job structure parent id also comes under payload

GowthamShanmugam:
@anivargi @brainfunked 
Messages are stored in different places 
1. If cluster_id present
     /clusters/cluster_id/Messages
else
    /nodes/node_id/Messages
2. If job_id present 
   stored in /Messages/jobs 
   also /Messages/events
else:
    only /Messages/events
 is this ok (or) job_updates should be stored in /Messages/jobs only not in /Messages/events, /clusters/cluster_id/Messages or /nodes/node_id/Messages.

which is correct?

PR #95, opened 2016-12-20
API integration to generate the hosts list with the monitoring data

anivargi:
Tendrl API needs to provide the storage, cpu usage, memory and alert
count per host in the hosts list API

### DIFF
--- a/specs/integrate_with_monitoring_api.adoc
+++ b/specs/integrate_with_monitoring_api.adoc
@@ -1,0 +1,107 @@
+// vim: tw=79
+
+= Integrate with monitoring API to generate hosts lists with monitoring data
+
+== Introduction
+
+Tendrl API needs to provide the storage, cpu usage, memory and alert
+count per host in the hosts list API
+
+== Problem description
+
+Tendrl API currently provides just the basic details like the fqdn, node_id and
+machine_id, these details are not sufficient and hence we need to add the
+monitoring details to the response per host.
+
+== Use Cases
+
+Provide a list of hosts to the user with details like storage, cpu usage,
+memory and alert count per host to generate the hosts list API
+
+== Proposed change
+
+Tendrl API needs to integrate with the monitoring API and find the details per
+host and associate them to the details fetched from etcd. These details help us
+generate the API for hosts listing.
+
+=== Alternatives
+
+None
+
+=== Data model impact
+
+None
+
+=== Impacted Modules
+
+==== Tendrl API impact:
+
+The proposed change will affect the hosts list API, the API will have more
+details about each host, the existing keys from the API are not affected and
+hence any service which integrated with the existing API will not be affected.
+
+==== Notifications/Monitoring impact:
+
+None
+
+==== Tendrl/common impact:
+
+None
+
+==== Tendrl/node_agent impact:
+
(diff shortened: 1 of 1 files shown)


anivargi:
@brainfunked @nthomas-redhat please review.

kamleshverma1:
@anivargi @shtripat @nthomas-redhat @brainfunked @gnehapk 
Ideally 'key' of Object should be some "static name" in order to access the value of it . it should not be dynamic (in this case - node_id) as such in response format -

Example: 
{

    "nodes":[
        {
            "3038c577-b233-4513-926d-a8a1ac63b0a3":  {

                   "fqdn":"dhcp43-189.lab.eng.blr.redhat.com",
                   "machine_id":"a2e4163a1a8e4b28afb1a17e6a1d9962",
                   "node_id":"3038c577-b233-4513-926d-a8a1ac63b0a3"

            },
            "stats":{}
        }
    ]

}

Here , i am fine with "stats" key which is static name. and also we can access the value of it using dot operator . but for "3038c577-b233-4513-926d-a8a1ac63b0a3"  , it can have any node-id which is dynamic (we don't know what it can have because it is not static name). and accessing this value in UI will be tricky and difficult . Can we have some static name instead of dynamic id?

Ideally it should be like this : 

Example: 
{

    "nodes":[
        {
            "detail":  {

                   "fqdn":"dhcp43-189.lab.eng.blr.redhat.com",
                   "machine_id":"a2e4163a1a8e4b28afb1a17e6a1d9962",
                   "node_id":"3038c577-b233-4513-926d-a8a1ac63b0a3"

            },
            "stats":{}
        }
    ]

}


Here all keys ( object properties name ) is static name .. not dynamic key like node-id.
Actually in UI , we should have well known key (property name) for any object to get that value . i don't know about back-end how to access this data . but in UI we should have well know key for object in order to access that data . 
But anyway we have work-around for this in javascript . but first i am thinking about if it is possible to show the data into UI directly without doing any logic on response data.

kamleshverma1:
@anivargi I observed that 'percent-used' (property name) is coming for all stats data (storage, cpu, memory) **But ideally property name of any object should not contain any special characters in JavaScript.**
Here i found that 'percent-used'  which contain '-' hyphen , which is not acceptable by JavaScript object.
Can we have underscore instead of this ?

gnehapk:
@anivargi Can we have multiple pools associated with a ceph cluster? If yes then the "pools" attribute in the clusterList.json should be an array instead of object. Please correct me if I am wrong.

gnehapk:
@anivargi We cant see any data for "RBD" in clusterList.json response. Can you please check whether it is there with some other name?

mkudlej:
Please add https://github.com/Tendrl/usmqe-tests/issues/15 to reference of testing.

kamleshverma1:
@anivargi @brainfunked @nthomas-redhat @shtripat i need API response format for GetNodeList which will contain info about cluster to be import.

anivargi:
@gnehapk The RBD' details will be automatically available when its populated in etcd, I unfortunately don't know the attributes and the structure of how it will be represented in GetClustersList response.

@kamleshverma1 I am working on adding the details.

Issue #44, opened 2016-12-09
Node agent tagging and tag based job routing


brainfunked:
It should be possible to assign tags to individual node agents. Flows could have tags statically indicates so that once jobs are created from these flows, the tags can be included in the jobs' metadata. Node agents with the matching tag should prioritise on picking up these jobs.

In the future this functionality will allow priority based job routing or role based job routing. For instance, if a node agent has been tagged as the provisioning controller, it will pick up the jobs tagged as provisioning. This is intended to be a framework feature to allow any kind of mapping, however the mapping itself would be introduced in a static manner in the code by the developers, depending upon the evolution of the functionality of the node agents.

This would impact the following:
* The node_context object would contain a list of tags,
* The definition schema would be updated to allow flows to contain preferred tags,
* The API, validation etc. need to accommodate the tags being present in the flows and being added as metadata to the jobs.

nnDarshan:
**NODE AGENT TAGGING:**
As first phase of implementation on node agent tagging, I have considered tags needed for basic import cluster. they are as follows:
* TENDRL - All nodes where tendrl-node-agent service is running will be tagged as TENDRL node
* TENDRL_SERVER - Node where tendrl-api and etcd service are running will be tagged with this tag.
* TENDRL_GLUSTER - Nodes where tendrl-gluster-integration service is running will be tagged with this tag.
* TENDRL_CEPH - Nodes where tendrl-ceph-integration service is running will be tagged with this tag.
* GLUSTER - Nodes where glusterd service is running
* CEPH - Nodes where ceph-osd or ceph-mon service is running
* CEPH_OSD - Nodes where ceph-osd service is running
* CEPH_MON - Nodes where ceph-mon service is running 

NOTE: Tags related to monitoring and provisioning will be added later

nnDarshan:
@brainfunked @shtripat @nthomas-redhat @r0h4n @anmolbabu Please provide your thoughts regarding above comment

shtripat:
@nnDarshan The only point I have is related to TENDRL_GLUSTER and GLUSTER or TENDRL_CEPH ad CEPH. How one is different from other? All the gluster nodes should have gluster-integration bits running.

nnDarshan:
@shtripat I feel this distinction is needed because, untill cluster is imported, gluster integration service wont be running. So a flow like ImportGlusterCluster can be targetted at GLUSTER nodes. But a flow like GlusterVolumeCreate should be targeted at TENDRL_GLUSTER node.

[...]

PR #98, opened 2016-12-22
Added spec monitor_ceph_pool_utilization.adoc

shtripat:

### DIFF
--- a/specs/monitor_ceph_pool_utilization.adoc
+++ b/specs/monitor_ceph_pool_utilization.adoc
@@ -19,13 +19,14 @@ pools.
 == Proposed change
 
 * Introduce a collectd plugin which collects the ceph pool utilization data from
-underlying cluster and pushes it to time series DB at regular intervals.
+underlying cluster
 
-* Enhance the monitoring module to set required thresholds for the ceph pool
-utilization
+* Introduce a template to configure the above mentioned collectd plugin with
+details such as cluster name and thresholds
 
-* Enhance the monitoring module to frame and raise required alerts while breach
-of threshold limits
+* Introduce threshold handlers on the alerting module to custom handle the
+pool(and cluster as well as ceph df itself provides both) utilization threshold
+breach
 
 === Alternatives
 
@@ -94,8 +95,7 @@ None
 
 ==== Tendrl API impact:
 
-* API module to support the fetching of ceph pool utilization data as well as
-part of listing generic monitoring data using resources
+None
 
 ==== Notifications/Monitoring impact:
 
@@ -120,13 +120,13 @@ None
 
 === Other end user impact:
 
-* Th trending graphs of ceph pool utilization data would help administrators to
+* The trending graphs of ceph pool utilization data would help administrators to
 decide on bringing in additional storage capacity in system based on warning and
 critical threshold breaches depicted clearly in Tendrl UI
 
 === Performance impact:
 
-* As confirmed by eph team running of `ceph df` command on mon nodes doesnt have
+* As confirmed by ceph team running of `ceph df` command on mon nodes doesnt have
 any major impact as its not heavy and fetches details from in memory mon details
 
 === Other deployer impact:
@@ -153,7 +153,8 @@ ceph mon use this intervals for fetching details from underlying cluster. So in
 ideal situation the interval for this plugin should be anything but more than 5
 seconds
 
-* Configure this plugin only if monitoring feature is enabled for Tendrl
+* Configure this plugin only if monitoring feature is enabled for Tendrl while
+cluster creation.
 
 * Deploy the plugin only on the mon nodes
 
--- a/specs/monitor_ceph_pool_utilization.adoc
+++ b/specs/monitor_ceph_pool_utilization.adoc
@@ -161,7 +161,7 @@ cluster creation.
 === Assignee(s):
 
 Primary assignee:
-  TBD
+  * anmolbabu
 
 Other contributors:
   * anivargi - API module
@@ -188,4 +188,8 @@ None
 
 == References:
 
-*
+* https://github.com/Tendrl/ceph_integration/issues/64
+
+* https://github.com/Tendrl/performance_monitoring/issues/14
+
+* https://github.com/Tendrl/alerting/issues/16
--- a/specs/monitor_ceph_pool_utilization.adoc
+++ b/specs/monitor_ceph_pool_utilization.adoc
@@ -1,0 +1,190 @@
+= Collect and process ceph pool utlization trending data
+
+The performance monitoring module if enabled in Tendrl, should collect and
+process the ceph pool utilization data in time series DB. This would help in
+creating trending charts in Tendrl UI which can help storage administrator to
+decide on requirement for additional storage in the system.
+
+== Problem description
+
+If ceph pool utilization trending is not available to administrator, they cannot
+easily decide how the pools are getting filled up and at what point in time more
+additional storage to be brought in and provisioned in the system.
+
+== Use Cases
+
+This PR addresses the trending charts use case in Tendrl for the ceph storage
+pools.
+
+== Proposed change
+
+* Introduce a collectd plugin which collects the ceph pool utilization data from
+underlying cluster and pushes it to time series DB at regular intervals.
+
+* Enhance the monitoring module to set required thresholds for the ceph pool
+utilization
+
+* Enhance the monitoring module to frame and raise required alerts while breach
+of threshold limits
+
+=== Alternatives
+
+* There are different ways to achieve the reading of pool utilization data
+
+* Option-1: Use rados module directly to invoke mon_command `df` to fetch the
+pool utilization data. A connection to rados could be kept open and keep running
+`df` command for getting details
+
+```
+import time
+import rados
+import json
+
+def get_pool_utilization(r):
+    cmd = {"prefix":"df", "format":"json"}
+    ret, buf, errs = r.mon_command(json.dumps(cmd), b'', timeout=5)
+    result = json.loads(buf)
+    return result
+
+try:
+    r = rados.Rados(conffile="/etc/ceph/ceph.conf")
+    r.connect()
+
+    for i in range(1000):
+        print(get_pool_utilization(r))
+        time.sleep(1)
+except KeyboardInterrupt:
+    pass
+finally:
+    if r is not None:
+        r.shutdown()
+```
+
+* Option-2: The other option is to use execute command `ceph df` as a process
+and parse the output for data
+
+```
+import json
+import subprocess
+
+cluster = 'my_cluster'
+cmd = [
+    'ceph',
+    '--cluster',
+    cluster,
+    '-f',
+    'json-pretty',
+    'df'
+]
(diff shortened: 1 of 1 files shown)


specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,190 @@
...
+This PR addresses the trending charts use case in Tendrl for the ceph storage
+pools.
+
+== Proposed change
+
+* Introduce a collectd plugin which collects the ceph pool utilization data from
+underlying cluster and pushes it to time series DB at regular intervals.
+
+* Enhance the monitoring module to set required thresholds for the ceph pool
+utilization
```
anmolbabu: Proposed changes can be:

* Introduce a collectd plugin which collects the ceph pool utilization data from underlying cluster.
* Introduce a template to configure the above mentioned collectd plugin with details such as cluster name and thresholds.
* Introduce threshold handlers on the alerting module to custom handle the pool(and cluster as well as ceph df itself provides both) utilization threshold breach.

----
Note:
* Any stat that is collected by collectd is guaranteed to land in time series db by the collectd plugin configured as write plugin(in our case the one configured to write to time series db).
* The collectd plugin to capture pool utilization will be configured on the leader mon node in the cluster at the time of creation of cluster. The removal of pool does not require the removal of configuration of pool utilization plugin as it is auto handled by ceph in case of using ceph df command. However the removal of cluster MIGHT require removal of respective collectd configuration.
----

shtripat: Ok. will change accordingly

anmolbabu: Also I think it would be better to mention the command or approach taken in the plugin to collect usage stats because, as you have mentioned below there are many techniques available to achieve this task we probably need to evaluate the flaws of each of them and make a final call. 1 problem that I have seen with ceph df with json format option is it provides only used size and % used but not total. I remember hearing that it'll be easier to add it on ceph side but we might need to raise an issue and track it if we go with that approach.

shtripat:
@anmolbabu can you have a look at the https://github.com/rochaporto/collectd-ceph/tree/master/plugins
This plugin as well uses json output and pushes the details to time series DB.
I understand that in case of json we would have to calculate the percentage usage value but with suggestion from ceph team on https://bugzilla.redhat.com/show_bug.cgi?id=1360230, isnt it doable.

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+None
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* API module to support the fetching of ceph pool utilization data as well as
+part of listing generic monitoring data using resources
+
```
anmolbabu, line 99: Probably, this needs to be either of the below:

1. The pool object in tendrl ceph integration definitions, will provide a way to indicate a way to read the utilization data from some directory in etcd if monitoring is enabled and the tendrl-api will look into this path to provide the utilization data
or
2. The performance-monitoring application will periodically update the utilization data to a location in etcd and also the performance monitoring application will expose apis to access this data and the tendrl-api will proxy to it.

I personally think that approach 2 although is a 2-step process, is safer and better than approach 1 of involving changes to definitions on the fly once the monitoring is enabled.

shtripat:
>Probably, this needs to be either of the below:
>
>    The pool object in tendrl ceph integration definitions, will provide a way to indicate a way to read the >utilization data from some directory in etcd if monitoring is enabled and the tendrl-api will look into this path >to provide the utilization data
>    or
>    The performance-monitoring application will periodically update the utilization data to a location in etcd >and also the performance monitoring application will expose apis to access this data and the tendrl-api will >proxy to it.

>I personally think that approach 2 although is a 2-step process, is safer and better than approach 1 of >involving changes to definitions on the fly once the monitoring is enabled.

Will change the spec as suggested

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+None
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* The performance monitoring application will periodically update the utilization
+data to a location in etcd and also the prformance monitoring application will
+expose APIs to access this data and the tendrl-api will proxy to it.
+
```
anmolbabu, line 99: As we discussed this goes as part of the pool entity itself. And since the update of utilization is anyway being done as part of pool entity and is regularly synced to etcd, by the ceph-integration, we no more require this api from performance-monitoring

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+None
+
+=== Security impact:
+
+None
+
+=== Other end user impact:
+
+* Th trending graphs of ceph pool utilization data would help administrators to
```
anmolbabu: Typo here: 'The'

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+decide on bringing in additional storage capacity in system based on warning and
+critical threshold breaches depicted clearly in Tendrl UI
+
+=== Performance impact:
+
+* As confirmed by eph team running of `ceph df` command on mon nodes doesnt have
```
anmolbabu: Typo here : 'ceph' ?

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+* Out of two alternatives both are equally good, but in this scenario usage of
+option-2 would be more meaningful
+
+* Make sure interval for running this plugin is nothing less than 5 seconds as
+ceph mon use this intervals for fetching details from underlying cluster. So in
+ideal situation the interval for this plugin should be anything but more than 5
+seconds
+
+* Configure this plugin only if monitoring feature is enabled for Tendrl
```
anmolbabu: "And the plugin will be configured once a cluster is created And removal of pools has no impact on the plugin or its configuration. However, removal of the cluster will handle the removal of configuration of this plugin. The details regarding cluster removal will however be covered in a separate spec as it is beyond the scope of this spec."

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+* API module owner(s) to take care of any API impacts
+
+* Alerting module owner(s) to take care handling of threshold breaches for ceph
+pool utilization
+
+== Implementation:
+
+* Add a collectd plugin which runs `ceph df` command at regular interval and
+pushes the data to time series DB
```
anmolbabu, line 146: 1 problem that I have seen with ceph df with json format option is it doesn't provide %used details and total capacity of pool. I remember hearing that it'll be easier to add it on ceph side but we might need to raise an issue and track it if we go with that approach. Also in normal cli command without json o/p format, it provides used percentage but not total but we'll need to parse the command output as a string.

specs/monitor_ceph_pool_utilization.adoc
shtripat: ok got it. will remove this

specs/monitor_ceph_pool_utilization.adoc
shtripat: Done

specs/monitor_ceph_pool_utilization.adoc
shtripat: Done

specs/monitor_ceph_pool_utilization.adoc
shtripat: I rember in skyring it was done using the non-json output. Will try to the same thing.

specs/monitor_ceph_pool_utilization.adoc
shtripat: ok

brainfunked (commented):
Looks good, will approve once the changes requested by Anmol have been made.

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,192 @@
...
+option-2 would be more meaningful
+
+* Make sure interval for running this plugin is nothing less than 5 seconds as
+ceph mon use this intervals for fetching details from underlying cluster. So in
+ideal situation the interval for this plugin should be anything but more than 5
+seconds
+
+* Configure this plugin only if monitoring feature is enabled for Tendrl
+
+* Deploy the plugin only on the mon nodes
```
anmolbabu, line 159: The plugin needs to be configured only on the leader mon node(only 1 node) because if it it is running on multiple nodes of a cluster, all of them will raise threshold breach alerts(if any) for the same set of pools in cluster and hence there will be as many duplicate alerts as the number of mons in cluster(if deployed on all mons) which can be handled internally(if un-avoidable) but it might be inefficient.

specs/monitor_ceph_pool_utilization.adoc
shtripat: We should be using non json output of `ceph df` command and use here. I feel @anmolbabu already has a patch for this.

specs/monitor_ceph_pool_utilization.adoc
shtripat: We are thinking of having some kind of tagging mechanism for the nodes and using that the leader mon can have this enabled.
@brainfunked what do you feel? Should we keep this requirement here at the moment or later we would take care.

specs/monitor_ceph_pool_utilization.adoc
anmolbabu: @shtripat I have a patch for collectd plugin @https://github.com/Tendrl/ceph_integration/pull/11

specs/monitor_ceph_pool_utilization.adoc
```diff
@@ -0,0 +1,191 @@
...
+Primary assignee:
+  TBD
+
+Other contributors:
+  * anivargi - API module
+  * anmolbabu - alerting module
+
+=== Work Items:
+
+* https://github.com/Tendrl/specifications/issues/97
```
anmolbabu, line 172: I think you probably need to create issues on associated components and link it here not the spec issue(I feel)

anmolbabu (commented):
1 minor comment and 1 answer awaited from @brainfunked

specs/monitor_ceph_pool_utilization.adoc
shtripat: Different module specific issues actually I keep under references and spec issue here :) thats what was followed in earlier specs.

anmolbabu: ok can you please add the component issues under References then

shtripat: Done

anmolbabu (approved):
LGTM but awaiting @brainfunked responses on "We are thinking of having some kind of tagging mechanism for the nodes and using that the leader mon can have this enabled.
@brainfunked what do you feel? Should we keep this requirement here at the moment or later we would take care"

shtripat:
@anmolbabu https://github.com/Tendrl/specifications/issues/44 talks about this tag based job execution and its linking with role based job execution on node agents.

Issue #87, opened 2016-12-19
Namespaces and flows to be carried with the node agent by default


brainfunked:
Tendrl API discovers and advertises any possible functionality via the namespaced flows in the definition files. This approach allows for addition flows to be implemented and made available dynamically via the API.

Node agent serves as the component responsible for implementing the flows and namespaces that provide the core tendrl functionality. The core tendrl functionality such as service detection, hardware inventory etc. is served via hardcoded flows and code shipped with the node agent. Wherever applicable, the core functionality is enabled via commons, as a framework feature.

Building on the core framework, support for specific storage system features of pluggable stacks can be enabled, without impacting the core framework itself. However, in some cases, it is impossible to achieve a complete separation between the supported storage systems and Tendrl. For example, to enable support to detect (https://github.com/Tendrl/specifications/issues/46), import (https://github.com/Tendrl/specifications/issues/54) and provision Ceph (https://github.com/Tendrl/specifications/issues/48) and Gluster (https://github.com/Tendrl/specifications/issues/49), the node agent itself needs to ship with at least the essential glue that could pull in the necessary integration components dynamically (https://github.com/Tendrl/specifications/issues/47, https://github.com/Tendrl/specifications/issues/51, https://github.com/Tendrl/specifications/issues/52).

This specification has been created to clearly separate out the various integrations by shipping only the bare necessary namespaces and flows that would ship with the node agent itself to allow the addition components to be then pulled in dynamically and in the future allow the support for different versions of the storage systems or even new storage systems to be added dynamically to a Tendrl deployment without requiring any changes in the code.

A part of this specification had originally been intended to be documented in the specification to define the scope of the integration modules, https://github.com/Tendrl/specifications/issues/52. Once this specification has been accepted, the details will be populated in that issue accordingly.

In addition to storage system support, the pluggable tendrl stacks such as monitoring, need to be allowed to be dynamically plugged into an existing Tendrl deployment. The tendrl namespace needs to allow support for such additional definition file updates to be deployed. This specification would also address this problem. It should enable the provisioning of the monitoring stack, https://github.com/Tendrl/specifications/issues/42, to progress.

r0h4n:
Apart from provisioning the mon and alerting stack via the node-agent, we have implemented the rest

PR #100, opened 2016-12-26
(WIP) Node agent system detection and integration modules

brainfunked:
Against https://github.com/Tendrl/specifications/issues/87.

Currently details the node agent design wrt the system detection modules and the integration modules.

### DIFF
--- a/specs/node_agent_flows_and_namespaces.adoc
+++ b/specs/node_agent_flows_and_namespaces.adoc
@@ -149,3 +149,332 @@ Every pluggable stack would need to supply it's own service detection,
 provisioning and, if applicable, integration modules. Each of these modules
 need to add the appropriate flows.
 
+
+== Data Model Impact
+
+The structure of the definition files would need to be updated to support
+provisioning and auto-detection functionality.
+
+A master definition file needs to be created, which would contain the essential
+global Tendrl namespace. Following is a commented version of what such a master
+definition file would contain. The example below sources all the definitions
+from a single file. However, this could be split up in multiple files and the
+namespaces combined as part of the centralised compilation process. The
+compilation process would be covered by specification
+https://github.com/Tendrl/specifications/issues/39[#39].
+
+  # The global tendrl namespace would consist mostly of hardcoded keys and
+  # values. Most of these keys and values would be reused in all the other
+  # definition files. The code associated with this namespace would also
+  # contain specific logic to handle the declarations in this section and the
+  # corresponding references from the other namespaces of the definition files.
+  namespace.tendrl:
+    # Installation sources define the supported installation methods and the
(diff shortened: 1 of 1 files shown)


specs/node_agent_flows_and_namespaces.adoc
```diff
@@ -0,0 +1,151 @@
+// vim: tw=79
+:toc:
```
shtripat, line 2: The only point I have at the moment is that it looks very high level to me and with more examples and snippets added from definitions would make it clearer for everybody else.

specs/node_agent_flows_and_namespaces.adoc
```diff
@@ -0,0 +1,623 @@
...
+
+* By default, two types of integration modules need to be shipped with the node
+  agent:
+** Service detection modules, which have been covered in the <<Service Detection Modules>> section.
+** Provisioning modules to deploy the integration bridges or provisioning
+   systems (gdeploy, ceph-ansible etc.) for a specific storage system, which
+   have been covered in the <<Provisioning Modules>> section.
+* For both types of modules, node agent needs to define a separate directory
+  from where the specific type of integration provisioning modules can be loaded.
+* The modules can be python classes with a specific interface method to be
```
shtripat, line 98: I feel this spec should list wire-frame of these interfaces in implementation section.

specs/node_agent_flows_and_namespaces.adoc
```diff
@@ -0,0 +1,623 @@
...
+  https://github.com/Tendrl/specifications/issues/49[#49].
+* Various provisioning flows,
+  https://github.com/Tendrl/specifications/issues/47[#47],
+  https://github.com/Tendrl/specifications/issues/51[#51],
+  https://github.com/Tendrl/specifications/issues/52[#52].
+
+
+== Proposed Changes
+
+=== Platform Detection Modules
```
nnDarshan, line 66: will there be multiple modules to detect platform ? cant this be done by a single module if its just detection of the OS and its version ?

specs/node_agent_flows_and_namespaces.adoc
```diff
@@ -0,0 +1,623 @@
...
+            # Provide either a list of success or failure exit codes. Anything
+            # else would indicate the other condition.
+            success_exit_code: Array
+      module:
+        # A specific module in the code
+        parameters:
+          - module: Module
+            parameters: Hash
+    storage_system_types:
+      ceph: namespace.tendrl.integrations.ceph
```
r0h4n, line 248: we call it "namespace.tendrl.ceph_integration", i suppose we have to change it to "namespace.tendrl.integrations.ceph"

same for gluster

r0h4n (changes requested):

PR #104, opened 2016-12-28
Added spec multiple_node_ids_in_jobs.adoc

shtripat:

### DIFF
--- a/specs/multiple_node_ids_in_jobs.adoc
+++ b/specs/multiple_node_ids_in_jobs.adoc
@@ -1,0 +1,119 @@
+= Pass multiple node-ids in a job submitted to central store job queue
+
+Currently a job can be targeted to a specific node by explicitly setting the
+attribute `node_id` in job structure. The matching the node-id picks up and
+processes the job from central store job queue.
+
+We should be able to submit job targeted to a set of nodes rather submitting to
+a specific node. This feature would implicitly enable targeting of a job to
+specific node anyway.
+
+== Problem description
+
+The job targeted to a single node restricts user and also it becomes a single
+point of failure. Additionally in case of `ceph` certain jobs can be and should
+be picked of any of the mon nodes in the cluster and we should not be
+restricting to a specific node.
+
+User should be able to pass set of node-ids in job and any one of the storage
+nodes underlying should pick and process this job.
+
+== Use Cases
+
+* Job execution by storage nodes from central store job queue
+
+== Proposed change
+
+* Enhance the job execution modules in Tendrl to looks for the job-ids list from
+the fetched job out of central store job queue, and process only if current
+node's id matches one in the list
+
+=== Alternatives
+
+None
+
+=== Data model impact:
+
+* Update the job definition to include an optional node-ids list rather than a
+specific node-id (sample job)
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* If API decides to target the job to a specific set of nodes, it should set the
+`node_ids` attribute for the job while submitting to job queue as a list of node
+ids
+
+==== Notifications/Monitoring impact:
+
+* None
+
+==== Tendrl/common impact:
+
+* Enhance the rpc job process handler to look for the `node_ids` attribute and
+make sure if the current node id falls under the list then only pick and process
+the job
+
+==== Tendrl/node_agent impact:
+
+None
+
+==== Sds integration impact:
+
+None
+
+=== Security impact:
+
+None
+
+=== Other end user impact:
+
+None
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+None
+
+=== Developer impact:
+
+None
+
+== Implementation:
+
+* Enhance the rpc job process handler in common module to look for the
+`node_ids` attribute and of the current node's id fals under this list then only
+pick and process the job
+
+* Update the sample job structure to include `node_ids` attribute
+
+=== Assignee(s):
+
+Primary assignee:
+  shtripat
+
+Other contributors:
+
+=== Work Items:
+
+* https://github.com/Tendrl/specifications/issues/102
+
+== Dependencies:
+
+None
+
+== Testing:
+
+* Verify if a job targeted at specific nodes is picked by one of them only
+
+== Documentation impact:
+
+None
+
+== References:
+
+* https://github.com/Tendrl/common/issues/86
--- a/specs/multiple_node_ids_in_jobs.adoc
+++ b/specs/multiple_node_ids_in_jobs.adoc
@@ -26,7 +26,11 @@ nodes underlying should pick and process this job.
 
 * Enhance the job execution modules in Tendrl to look for the `node_ids` list from
 the fetched job out of central store job queue, and process only if current
-node's id matches one in the list
+node's id matches one in the list.
+
+This change still supports targeting the job to a specific node by setting only one node-id
+in parameter `node_ids`. If the job is to be picked by any node simply just dont set the
+parameter `job_ids` in job.
 
 === Alternatives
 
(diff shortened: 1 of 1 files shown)


specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
+= Pass multiple node-ids in a job submitted to central store job queue
+
+Currently a job can be targeted to a specific node by explicitly setting the
+attribute `node_id` in job structure. The matching the node-id picks up and
```
anmolbabu, line 4: "The node-agent on the node whose id matches the passed node_id picks up...." ?

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+processes the job from central store job queue.
+
+We should be able to submit job targeted to a set of nodes rather submitting to
```
anmolbabu: rather than?

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+a specific node. This feature would implicitly enable targeting of a job to
+specific node anyway.
+
+== Problem description
+
+The job targeted to a single node restricts user and also it becomes a single
+point of failure. Additionally in case of `ceph` certain jobs can be and should
+be picked of any of the mon nodes in the cluster and we should not be
```
anmolbabu, line 15: "picked by any of the....".

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+== Use Cases
+
+* Job execution by storage nodes from central store job queue
+
+== Proposed change
+
+* Enhance the job execution modules in Tendrl to looks for the job-ids list from
+the fetched job out of central store job queue, and process only if current
+node's id matches one in the list
+
```
anmolbabu, line 34: But how can this handle the case of "any one of the" as briefed in the "Problem Description" section in "User should be able to pass set of node-ids in job and any one of the storage nodes underlying should pick and process this job."

I feel the case of "any one of the" should be made by the application loading the job and the case of execute by one/or set of nodes will be handled as you described in "Proposed Change". Is my understanding correct?

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+* https://github.com/Tendrl/specifications/issues/102
+
+== Dependencies:
+
+None
+
+== Testing:
+
+* Verify if a job targeted at specific nodes is picked by one of them only
```
anmolbabu, line 115: one or all of them?

specs/multiple_node_ids_in_jobs.adoc
shtripat: yes. the current node's id should be in the list passed in the job

specs/multiple_node_ids_in_jobs.adoc
shtripat: Done

specs/multiple_node_ids_in_jobs.adoc
shtripat: Done

specs/multiple_node_ids_in_jobs.adoc
shtripat: Thats correct. So say a job has passed node_id1/2/3 as `node_ids` in that case any of the nodes with node_id1/2/3 would pick this job and process. May be I would rephrase the sentence a bit.

specs/multiple_node_ids_in_jobs.adoc
shtripat: It should be picked by one of them and not all. Once a job is picked by one node, it would not be picked by others in the list. If you a job to be picked by set of nodes, you need to create those many jobs with the specific node_id as single entry in `node_ids` field of the jobs.

anmolbabu (commented):
The only concern I see is what happens if more than 1 node-agent's pick up the job simultaneously and both of them try to update the job status simultaneously, is there a locking mechanism to lock an update to the job in etcd so that the other can be aware that someone has picked it up?

Otherwise Looks good to me

mkudlej:
I don't think that this change requires any special testing. Everything should work properly so this change will be tested only partially by API calls. 
Do you agree? @mbukatov @fbalak @dahorak @ltrilety

shtripat:
@anmolbabu regarding locking yes there is a way to lock the item while processing. I feel there is a plan to do the same while picking any job but not as part of this PR.

anmolbabu:
@shtripat ok

anmolbabu (approved):
LGTM

nnDarshan:
LGTM

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+point of failure. Additionally in case of `ceph` certain jobs can be and should
+be picked of any of the mon nodes in the cluster and we should not be
+restricting to a specific node.
+
+User should be able to pass set of node-ids in job and any one of the storage
+nodes underlying should pick and process this job.
+
+== Use Cases
+
+* Job execution by storage nodes from central store job queue
```
r0h4n, line 23: Give a concrete flow example, I am not sure why we need this spec since we already round robin node agents to pickup jobs

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+the fetched job out of central store job queue, and process only if current
+node's id matches one in the list
+
+=== Alternatives
+
+None
+
+=== Data model impact:
+
+* Update the job definition to include an optional node-ids list rather than a
```
cybernth, line 41: Can you add the details about the job structure changes?
* current structure
* New fields if any added to the structure
*How I can target a single node/specific nodes/all nodes

specs/multiple_node_ids_in_jobs.adoc
```diff
@@ -0,0 +1,119 @@
...
+specific node-id (sample job)
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* If API decides to target the job to a specific set of nodes, it should set the
```
cybernth, line 48: Does this change backward compatible? Does the existing APIs require any change?

cybernth (changes requested):

specs/multiple_node_ids_in_jobs.adoc
shtripat: Yes it should work seamlessly as the parameter is not a mandatory one.

specs/multiple_node_ids_in_jobs.adoc
shtripat: >Can you add the details about the job structure changes?
>
>   current structure

As part of https://github.com/Tendrl/commons/pull/92 the sample job creator is changed already

>    New fields if any added to the structure
>    *How I can target a single node/specific nodes/all nodes

If the value `node_ids` contains only one node_id in list, the job would be picked by that specific node only. 
This change is about targetting the job to a set of nodes and only one of them would pick and process. If you want job to be picked by any node from the full list dont pass the parameter `node_ids` in job and any node from full list would pick and process.

specs/multiple_node_ids_in_jobs.adoc
shtripat: Yes old code should work seamlessly as the parameter is not a mandatory one.

specs/multiple_node_ids_in_jobs.adoc
cybernth: Got it. Can you add these specifics into the spec?

shtripat: Ok. Added a section above for the same

specs/multiple_node_ids_in_jobs.adoc
r0h4n: @shtripat send a PR to this spec with a flow example later, merging this now.

r0h4n (approved):

PR #105, opened 2016-12-28
spec for import cluster

gnehapk:


gnehapk:
Against #56

gnehapk:
@brainfunked @kamleshverma1 @nthomas-redhat @anivargi @shtripat Please review it.

specs/import-cluster-view.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+* On success, a successful message will be displayed to user.
+  (https://redhat.invisionapp.com/share/6T900V2ZX#/screens/198651334)
+
+== Proposed change
+
+Front-end related changes:
+
+	* Create views for Landing page (First Time Experience), Import cluster page and page where successful message will be displayed.
+	* Implement client side filer and sorting functionality using angular's filter and orderby components.
```
gnehapk: @julienlim What are the options for sorting and filtering on import cluster view - https://redhat.invisionapp.com/share/6T900V2ZX#/screens/198042641

mkudlej:
Please add https://github.com/Tendrl/usmqe-tests/issues/20 as reference for testing.

anivargi (approved):
Looks good.

specs/import-cluster-view.adoc
julienlim: @gnehapk 

The following are the options for sort and filter for the Import Cluster view.

Filter (by): name, IP address, role (if available for Ceph)
Sort (by): name, IP address, role (if available for Ceph)

shtripat: s/filer/filter/g

shtripat (commented):
LGTM. One minor spelling mistake.

specs/import-cluster-view.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+None
+
+=== Data model impact:
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+Not yet decided.
```
gnehapk: @anivargi Please provide the APIs and their response which will be used in Import cluster functionality.

specs/import-cluster-view.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+== Use Cases
+
+This addresses the use case of importing existing clusters(ceph. gluster etc.) into tendrl for management.
+
+The workflow is as follows:
+
+* "Import Cluster" can be triggered from 
+
+	- Landing page (First Time Experience) if no cluster is present in tendrl
```
cybernth: Does the landing page have this option if atleast one cluster is present? I assume it lands into the dashboard in that case

specs/import-cluster-view.adoc
```diff
@@ -0,0 +1,133 @@
...
+	  (https://redhat.invisionapp.com/share/6T900V2ZX#/screens/198042644)
+
+	- Cluster List View if minimum one cluster is present in tendrl.
+	  (https://redhat.invisionapp.com/share/BR8JDCGSQ#/screens/185937524)
+
+* User clicks on "Import Cluster". 
+
+* It will take to host list page where user will select the hosts he wants to import.
```
cybernth: User selects a single host participating in a cluster. Upon clicking 'Next" the API gets a list of hosts participating in the cluster. I think we need to reword

specs/import-cluster-view.adoc
```diff
@@ -0,0 +1,133 @@
...
+  (https://redhat.invisionapp.com/share/6T900V2ZX#/screens/198651334)
+
+== Proposed change
+
+Front-end related changes:
+
+	* Create views for Landing page (First Time Experience), Import cluster page and page where successful message will be displayed.
+	* Implement client side filer and sorting functionality using angular's filter and orderby components.
+	* Send API call for host list whenevr users click "Refresh" button.
+	* Show total count of hosts which can be imported.
```
cybernth: What does this mean?

cybernth (changes requested):

specs/import-cluster-view.adoc
gnehapk: @nthomas-redhat When the application gets loaded, the 1st call we make is for get cluster list call. Based on its reponse, the app decides whether it should go to landing page or cluster list view.

gnehapk:
@nthomas-redhat @brainfunked @anivargi @shtripat @kamleshverma1 I have updated the PR. Please review.

kamleshverma1:
@julienlim As we(front-end guys) have discussed with Nishanth and mrugesh, we are not going to consider bootstrap node concept while importing cluster. All clusters will be auto detected at the time of import. and in case of multiple cluster , we will have some "select drop-down" from where user can select appropriate cluster which they want to be import. or You can suggest if there is any other alternative instead of using "select drop-down" in case of multiple cluster .

here i attached screen-shot for import cluster
screenshot from 2017-01-13 11 58 38

kamleshverma1:
@julienlim i want some clarification on primary navigation visibility while importing cluster .
i assumed that basically there is two case for import cluster 

1) If there is at least one cluster in Tendrl - primary navigation should be visible so that user can navigate into any other pages because we have at least one cluster in this case.

2) If there is no any cluster in Tendrl - primary navigation should be hidden because there is no point to show navigation as we don't have anything (clusters, pools, file shares etc ) in this case.

please correct me if i am wrong

kamleshverma1:
@anivargi I have gone through response format which you sent for nodelist and clusterlist .
and i observed that "stats" coming as string . 

"stats":"{'alert_cnt': 10, 'storage': {'total': 42949672960, 'used': 21474836480, 'updated_at': '2016-12-19T23:12:51.278616', 'percent_used': 50}, 'cpu': {'updated_at': '2016-12-19T23:12:51.278516', 'percent_used': 4.5}, 'memory': {'total': 4294967296, 'used': 2147483648, 'updated_at': '2016-12-19T23:12:51.278593', 'percent_used': 50}}"

Ideally it should be object instead of string in order to access the value of it . otherwise we need to do some extra logic unnecessary to get the value of it in UI.

Issue #75, opened 2016-12-15
Host, Cluster, File shares, RBDs and Pools List view


gnehapk:
- Make changes to host and cluster list view to match with UX design.
- Create view for File share, RBD and Pool List.
- Confirm the data to be shown for these list views from API team.

PR #108, opened 2016-12-29
specification for RBDs list view

kamleshverma1:
specification for RBDs list view

### DIFF
--- a/specs/rbd_list_view_component.adoc
+++ b/specs/rbd_list_view_component.adoc
@@ -1,0 +1,127 @@
+// vim: tw=79
+
+= Tendrl Definitions: Display list of RBDs present in Ceph cluster in Tendrl
+
+Display list of RBDs associated with a ceph cluster present in Tendrl. The list should have functionality for 
+
+* filtering by RBD name, cluster name and backing pool name.
+* sorting by RBD name, utilization, cluster name, status and backing pool.
+
+RBDs list will be sorted by name by default.
+
+== Problem description
+
+Tendrl should show list of RBDs associated with a ceph cluster along with their details.
+
+The details shown for each RBD in UI is - (https://redhat.invisionapp.com/share/BR8JDCGSQ#/screens/198417119)
+
+* RBD Name
+* Status
+* RBD Utilisation
+* Cluster Name
+* Backing pool with indicator if pool is shared with other RBDs.
+* Active alerts
+
+== Use Cases
+
+Users can see the list of RBDs associated with a ceph cluster present in Tendrl. They can also 
+
+* filter by RBD name, cluster name and backing pool name.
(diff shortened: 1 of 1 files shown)


kamleshverma1:
@gnehapk @brainfunked @anivargi @shtripat @nthomas-redhat Please review it.

kamleshverma1:
Addresses #75 .

specs/rbd_list_view_component.adoc
```diff
@@ -0,0 +1,127 @@
...
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+/api/1.0/GetClusterList - fetch the all detail along with RBDs list.
+
+Required from API - 
+
+* Response format for the API call
+* If "GetClusterList" contain RBDs list . than "parameter name" should be known which is  holding the RBDs list.
```
kamleshverma1: @anivargi @shtripat @brainfunked If "GetClusterList" contain RBDs list . so Can i have  "parameter name" which is  holding the RBDs list ? or Is there any other APi for RBDs list ?

shtripat: @kamleshverma1 I feel rbd listing should be a separate api like /clusters/{id}/GetRbdList. @anivargi can you confirm please?

mkudlej: I agree with @shtripat RBD list should be separated API call.

anivargi: @shtripat @mkudlej Currently the API responds with all cluster related objects in 1 API call like for ex: for Ceph we have pools, but we surely have individual API calls for all storage objects like pools, RBD's in the future.

@kamleshverma1 The info will be present in the GetClusterList API call, currently I am not aware of the structure of how the details are stored in etcd and what the attributes are called, so if @shtripat or @nthomas-redhat can have a instance which has the details populated then the API will automatically respond with the desired response.

kamleshverma1:
@gnehapk @cloudbehl @anivargi @nthomas-redhat @shtripat Please review this spec . i have updated this.

specs/rbd_list_view_component.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+* Name
+
+* Cluster Name
+
+* backing pool
+
+Required from API -
+
+* status
```
kamleshverma1, line 62: @anivargi @nthomas-redhat @shtripat according to UX mockup (https://redhat.invisionapp.com/share/BR8JDCGSQ#/screens/198417119) , missing these fields for RBDs in given response format by anup
1) status
2) utilization
3) alert count

shtripat: @kamleshverma1 `utilization` and `alert count` would be done as part of monitoring I feel. I dont see a `status` field as such for rbds in cli output from ceph.

PR #110, opened 2016-12-29
Component wise comparison for hardware inventory details

GowthamShanmugam:

### DIFF
--- a/specs/hardware_inventory_component_wise_check.adoc
+++ b/specs/hardware_inventory_component_wise_check.adoc
@@ -1,0 +1,119 @@
+= Component wise comparison for hardware inventory details
+
+Hardware inventory details needs component wise comparison.
+
+== Problem description
+
+* Currently, we are comparing hardware inventory details with entire file and
+  if it is not equal we are updating all the component details in etcd. Problem
+  here is if any changes in one component details then updation happens for all
+  the component.
+
+== Use Cases
+
+None
+
+== Proposed change
+
+* Instead of comparing hardware inventory details with entire file, Compare
+  component wise. It will reduce number of etcd request.
+
+== Alternatives
+
+* None
+
+== Data model impact
+
+None
+
+== Impacted modules
+
+=== Tendrl/common impact
+None
+
+=== Tendrl/node_agent impact
+
+* Modify the logic of comparing newly collected hardware inventory details with
+  entire file.
+* The comparison should happen component wise.
+  (more details in implementation section)
+
+=== Tendrl/gluster_integration impact
+
+None
+
+=== Tendrl/ceph_integration impact
+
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @brainfunked @r0h4n @nnDarshan please review this

specs/hardware_inventory_component_wise_check.adoc
```diff
@@ -0,0 +1,119 @@
...
+== Impacted modules
+
+=== Tendrl/common impact
+None
+
+=== Tendrl/node_agent impact
+
+* Modify the logic of comparing newly collected hardware inventory details with
+  entire file.
+* The comparison should be happen component wise.
```
shtripat: s/should be happen/should happen/g

shtripat (commented):
LGTM. 1 minor comment

specs/hardware_inventory_component_wise_check.adoc
GowthamShanmugam: i will change it

specs/hardware_inventory_component_wise_check.adoc
```diff
@@ -0,0 +1,119 @@
...
+None
+
+== Implementation
+
+* The comparison logic needs to be component wise comparison.
+* if any component details are not changed then delete that from hardware inventory
+  list. So only changed is passed to onpull function.
+```
+old_data = "from hardware inventory file"
+if new_data['component_name'] == old_data['component_name']:
```
shtripat, line 88: Are we comparing json strings here? We should convert to object and then compare. May be all the object should have a equal() function which returns true/false based on equality of the current object with other one.

GowthamShanmugam: @shtripat yes i am comparing json string... As per your comment i will convert to object and compare it

GowthamShanmugam: @shtripat we cant compare this as two different object, as per our implementation we cant create another object, actually we are sharing same object all the time.

shtripat: @GowthamShanmugam my only point if comparing json string as such I am not sure how foolproof it would be.

GowthamShanmugam: @shtripat yes, you are correct comparing json is not good, i will think any alternate way for this

GowthamShanmugam: I have disused with @shtripat , comparing json is not a problem so we can keep this approach for compare component

nnDarshan:
LGTM

shtripat:
LGTM

PR #116, opened 2017-01-10
Specification file for proposing changes after implementation of logging in commons(PR #94)

anmolsachan:
Signed-off-by: anmolsachan asachan@redhat.com

### DIFF
--- a/specs/logging_changes_spec.adoc
+++ b/specs/logging_changes_spec.adoc
@@ -1,0 +1,171 @@
+// vim: tw=avored-markdown/preview.ht79
+
+# Changes in tendrl modules after introduction of new logging module in common
+
+Since a new logging module is being introduced in Tendrl-common(https://github.com/Tendrl/specifications/pull/94), after its implementation
+there will be requirement of changes in all the modules of Tendrl where logging is used.
+This file proposes all those changes and how they will be made.
+
+
+## Problem description
+
+Many modules in Tendrl have a functionality of logging different types of situations/use cases such as logging (info, errors, warning etc.).
+Till now these were implemented by importing the Python standard library's logging module then calling logger functions to handle the logging.
+After implementation of new logging module there will be requirement of changes in all the modules of Tendrl where logging is used.
+
+#### Use Cases
+
+There will be 3 classes(use cases) that will be provided by the common logging module:
+
+ * Log - Will write logs to log files.
+ * Alert - Will send the alert messages to the node agent socket and write the same to the log files.
+ * Operations - Will write operation log to log files.
+
+#### Proposed change
+
+Earlier there was only provision of writing to log files, now this functionality in being diversified and extended.
+
+The following are the key points to be noted:
+
+* Introduction of alert and operations along with logs and all these will be treated as messages emitted by Tendrl modules.
+
+* There will 3 classes of messages that will be provided and will be used to achieve the required purpose:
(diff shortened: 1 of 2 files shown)


specs/logging_changes_spec.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+## Problem description
+
+Many modules in Tendrl have a functionality of logging different types of situations/use cases such as logging (info, errors, warning etc.).
+Till now these were implemented by importing the Python standard library's logging module then calling logger functions to handle the logging.
+After implementation of new logging module there will be requirement of changes in all the modules of Tendrl where logging is used.
+
+#### Use Cases
+
+There will be 3 classes(use cases) that will be provided by the common logging module:
```
shtripat: I feel use cases are logging, alerting and in flight operation steps and below classes are defined for that. We should not directly talk about classes. Rather we should say "three use cases which would be tackled are logging, alerting and in flight operations".

specs/logging_changes_spec.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+The following are the key points to be noted:
+
+* Introduction of alert and operations along with logs and all these will be treated as messages emitted by Tendrl modules.
+
+* There will 3 classes of messages that will be provided and will be used to achieve the required purpose:
+
+ ... Log - Will write logs to log files.
+ ... Alert - Will send the alert messages to the node agent socket and write the same to the log files.
+ ... Operations - Will write operation log to log files.
```
shtripat: And would also be persisted to central store (etcd) under task data

specs/logging_changes_spec.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+None
+
+#### Impacted Modules:
+
+#### Tendrl/common impact
+
+Similar changes like _example change_ mentioned above will be required in following modules :
+
+* /commons/etcdobj/etcdobj.py
```
shtripat: This is just list from commons module and would be required in other modules like node-agent, gluster-integration and ceph-integration, performance-monitoring and alerting etc. I feel you can just mention the module names.

specs/logging_changes_spec.adoc
```diff
@@ -0,0 +1,171 @@
...
+* /commons/utils/package_installer.py
+* /commons/utils/service.py
+* /commons/config.py
+* /commons/log.py
+
+#### Tendrl/node_agent impact
+
+Similar changes like _example change_ mentioned above will be required in following modules :
+
+* /node_agent/manager/manager.py
```
shtripat: Listing individual sources not required I feel and you just mention to replace all the logging instances with new framework.

specs/logging_changes_spec.adoc
anmolsachan: ok .. I will update the same

specs/logging_changes_spec.adoc
anmolsachan: ok .. I will update the same

specs/logging_changes_spec.adoc
anmolsachan: ok .. I will update

specs/logging_changes_spec.adoc
anmolsachan: ok .. I will update

PR #117, opened 2017-01-11
Specification for storage system detection

shtripat:

### DIFF
--- a/specs/sds_detection_modules.adoc
+++ b/specs/sds_detection_modules.adoc
@@ -1,0 +1,260 @@
+// vim: tw=79
+
+= SDS detection modules in node_agent
+
+Node agent serves as the component that is responsible for invoking and/or
+executing operations that are part of global Tendrl scope.
+
+Tendrl officially supports specific storage systems (ceph and gluster). However
+Tendrl also aims to be a framework that provides some essential capabilities
+and interfaces so that it is easy to add support for storage systems that are
+not officially supported.
+
+For this Tendrl should be able to auto-detect existing deployments and out of
+the box sane defaults for laying out the storage system and managing them.
+
+== Problem description
+
+Tendrl need to provide a pluggable model using which auto-detection modules
+for different storage systems can be deployed in pre-defined locations and
+Tendrl can load and execute certain generic APIs on these modules to figure the
+underlying pre-existing storage system. This can become the pre-requisite which
+gets executed as part of starting node agent on the storage nodes and helps
+import cluster flow in Tendrl to decide what kind of cluster is getting imported
+. This also would help deciding on deployment of corresponding Tendrl
+integration module for managing the cluster seamlessly.
+
+== Use Cases
+
+This specification covers the import cluster flow in Tendrl and auto detection
+of underlying storage systems on storage nodes.
+
+== Proposed change
+
+We try to introduce a pluggable set of modules (for officially supported storage
+systems from Tendrl) which could deployed in a standard Tendrl defined location
+and would be executed as part of node agent start on nodes. These modules would
+auto detect the underlying storage systems and its related details. These
+details would be tagged with storage nodes so that while import cluster Tendrl
+would be able to decide on deployment of corresponding integration modules on
+the storage systems.
+
+A summary of proposed changes are
+
+* Define a standard path from where all the available storage detection modules
+would be loaded
+
+* While starting the node agent service on the storage nodes, execute these
+plugins and auto detect the storage systems (if any) on underlying storage nodes
+
+* Tag the dectected details (if any) with the storage nodes details
+
+* Enhance the import cluster flows in node agent component to decide on what
+Tendrl integration module gets deployed on the underlying storage nodes
+
+=== Alternatives
+
+None
+
+=== Data model impact:
+
+None
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* API layer should attach the storage system specific auto-detected details of
+the nodes and list as part of node listing
+
+==== Notifications/Monitoring impact:
+
+None
+
+==== Tendrl/common impact:
+
+* Common module should define a generic interface for plugins which would
+be extended by different auto-detection modules specific interfaces. The
+structure of the generic interface could be as below
+
+```
+class AutoDetectPlugin(object):
+
+    @abstractmethod
+    def initialize(self):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def destroy(self):
+        raise NotImplementedError()
+```
+
+The class which takes care of mounting a plugin would be something like
+
+```
+class PluginMount(type):
+    def __init__(cls, name, bases, attrs):
+        if not hasattr(cls, 'plugins'):
+            cls.plugins = []
+        else:
+            cls.register_plugin(cls)
+
+    def register_plugin(cls, plugin):
+        instance = plugin()
+        cls.plugins.append(instance)
+        instance.initialize()
+```
+
+==== Tendrl/node_agent impact:
+
+* The node agent component need to define interfaces which extend from commons
+AutoDetectPlugin and add specific functions as below
+
+```
+class SDSDetectPlugin(AutoDetectPlugin):
+    @abstractmethod
+    def detect_storage_system(self):
+        raise NotImplementedError()
+```
+
+* The node agent component would introduce two python modules for ceph and
+gluster. These would be specific classes which extend from above SDSDetectPlugin
+and provide the actual implementation for the function
+
+```
+class DetectCephStorageSystem(SDSDetectPlugin):
+    def detect_storage_system(self):
+        # Implement the logic to get the details of underlying storage system
+        # 1. Figure out if ceph bits installed and set the type accordingly
+        # 2. Figure the version of ceph bits installed and accordingly set the
+        # version
+        # 3. Figure out the role of the storage node (by figuring out if MON or
+        # OSD processes are running) and accordingly set the role for the node
+        # 4. Generate (or read FSID in case of ceph) and assign a temporary
+        # cluster id. Write the this detail to a defined path as well on node
+```
+
+* Similary another implementation to be introduced for gluster as well
+
+* Enhance the node agent component to load these SDS detection plugins for pre-
+defined path (tendrl/node_agent/sds_detection/plugins) to mount them
+
+* While starting the node agent service execute the SDS detection plugins and
+tag the details to the storage node
+
+* If one plugin execution succeeds stop further execution
+
+==== Sds integration impact:
+
+* No impact as such on SDS integration modules due to auto detection and their
+role comes as usual for other management activities once cluster is imported in
+Tendrl system
+
+=== Security impact:
+
+None
+
+=== Other end user impact:
+
+* While import cluster flow, the underlying SDS details would be auto-detected
+and manifested in UI for displaying layout (nodes, roles, storage system version
+etc) of the cluster beforehand
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+None
+
+=== Developer impact:
+
+* API layer to take care of listing the tagged additional storage system details
+while listing the nodes for import cluster flow
+
+== Implementation:
+
+* The details of implementation are briefed in individual component impact
+sections already.
+
+* To summarize
+
+** Commons component to add a base interface for `AutoDetectPlugin`
+
+** Node agent component to extend the `AutoDetectPlugin` and define specific
+interface for SDS detection
+
+** Node agent to implement `SDSDetectPlugin` and add modules for `ceph` and
+`gluster`
+
+** Node agent to mount the SDS detection plugins while starting the service and
+execute these plugins one by one to figure out the details of the underlying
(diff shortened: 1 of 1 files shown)


shtripat:
@Tendrl/tendrl-core @Tendrl/tendrl_api please review

mkudlej:
Do I understand it right that after implementing this spec there is no need to add sds_version(and sds_name) parameter to API function calls?
You can find sds_version and sds_name as parameters here https://github.com/anivargi/documentation/blob/c26ccba15091efd70de41c54929f5f710626b8ec/api/ceph-pool-examples.adoc and https://github.com/Tendrl/documentation/blob/master/api/gluster-create-volume-example.adoc

If yes, then there is API impact and it should be mentioned in spec.

shtripat:
@mkudlej These values get tagged with storage nodes. I feel still jobs need to provide sds_type and sds_version fields fetched from nodes list while import cluster flows.

@brainfunked comments?

mkudlej:
@shtripat My question was not just about cluster import but more about all other functions for managing clusters and other cluster atoms.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+None
+
+==== Tendrl/common impact:
+
+* Common module should define a generic interface for plugins which would
+be extended by different auto-detection modules specific interfaces. The
+structure of the generic interface could be as below
+
+```
+class AutoDetectPlugin(object):
```
cybernth: why this is required?

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+* Tag the dectected details (if any) with the storage nodes details
+
+* Enhance the import cluster flows in node agent component to decide on what
+Tendrl integration module gets deployed on the underlying storage nodes
+
+=== Alternatives
+
+None
+
+=== Data model impact:
```
cybernth, line 56: Data model and tendrl defintions needs to be added

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+
+* Define a standard path from where all the available storage detection modules
+would be loaded
+
+* While starting the node agent service on the storage nodes, execute these
+plugins and auto detect the storage systems (if any) on underlying storage nodes
+
+* Tag the dectected details (if any) with the storage nodes details
+
+* Enhance the import cluster flows in node agent component to decide on what
```
cybernth: I think this is out of scope as far as this spec is concer?

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+
+=== Data model impact:
+
+None
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* API layer should attach the storage system specific auto-detected details of
```
cybernth: are you planning to do this as part of this spec. i feel limit the scope this spec only to detection of cluster details and updating the etcd

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+
+* SDS Detect plugin for `gluster`
+
+** Execute the command `gluster --version` to figure out if gluster bits are
+installed on the storage node. Also get the version details and tag with the
+node in central store
+
+** Default assign the role as `glusterd` for all the storage nodes
+
+** Execute the command `gluster pool list` and calculate as hash out of node
```
cybernth: Does this changes if the cluster is expanded after detecting the cluster?

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,260 @@
...
+UUIDs of the nodes in the trusted storage pool. Tag this as cluster_id for the
+nodes. Storage the details in pre-defined location (say
```
cybernth, line 259: I think https://github.com/Tendrl/specifications/pull/117 talks about detected_clusterand this information is supposed go there

cybernth (changes requested):

specs/sds_detection_modules.adoc
shtripat: Ok. will remove this.

specs/sds_detection_modules.adoc
shtripat: Done. will add how node entity would be tagged with additional data.

specs/sds_detection_modules.adoc
shtripat: Sure. Will remove from this spec.

specs/sds_detection_modules.adoc
shtripat: This is the base interface defined in commons module (with basic functions to initialize etc). The node agent would define SDSDetectPlugin interface extended from this (with abstract function detect_storage_system). This SDSDetectPlugin would be implemented by Ceph and Gluster specific classes which would work as detectors for underlying storage system.

I feel this is the same hierarchy we discussed as well earlier. Do you think something else?

specs/sds_detection_modules.adoc
shtripat: This would be used to create a unique UUID to identify the current cluster state before import. Once cluster imported into Tendrl, it could be kept as is and no need to change the cluster_id value based on newer/fewer nodes in the cluster at later stage.

@brainfunked  comments?

specs/sds_detection_modules.adoc
shtripat: Line 117 talk about CephSDSDetectionPlugin so I didnt add gluster specific details there. Here under *SDS Detect plugin for gluster* I have added detailed commands etc to be executed.

specs/sds_detection_modules.adoc
cybernth: I don't think we need a top level base class in common, only pluginmount is required here. Also Rohan pointed to a doc where some frameworks could help here. Good to refer that as well. --> http://docs.openstack.org/developer/stevedore/

shtripat: ok

shtripat: @r0h4n If we plan to use http://docs.openstack.org/developer/stevedore/ do we need to build some package from openstack? How this dependency would be brought in Tendrl?

For sure it looks like serving the purpose we need by using `extension.ExtensionManager` and we would be able to load of the extension available as plugins for detection.

specs/sds_detection_modules.adoc
```diff
@@ -49,47 +49,104 @@ plugins and auto detect the storage systems (if any) on underlying storage nodes
...
+          type: Create
+          uuid: eda0b13a-7362-48d5-b5ca-4b6d6533a5ab
+      attrs:
+        cmd_str:
+          type: String
+        fqdn:
+          type: String
+        status:
+          type: Boolean
+        cluster_id:
```
cybernth: I think this is not the right way. Don't add as attribute, rather do it as different object like node_context

cybernth (changes requested):

specs/sds_detection_modules.adoc
shtripat: I am ok with this but my understanding was that these details need to be tagged with node inventory.
@brainfunked @r0h4n @nnDarshan comments?

specs/sds_detection_modules.adoc
shtripat: As discussed we would go ahead with existing logic for loading the plugins dynamically and in later releases would try to use `stevedore` module for this.

specs/sds_detection_modules.adoc
shtripat: Changed the logic to tag the additional details as part of Node_context

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+```
+class NodeContext(EtcdObj):
+    __name__ = 'nodes/%s/Node_context'
+
+    node_id = fields.StrField("node_id")
+    machine_id = fields.StrField("machine_id")
+    fqdn = fields.StrField("fqdn")
+    cluster_id = fields.StrField("cluster_id")
+    sds_pkg_name = fields.StrField("sds_pkg_name")
```
brainfunked, line 68: Drop both sds_pkg_name and sds_pkg_version. Use only roles. Put up a global namespace for /roles. Each detection module has a hardcoded role that it assigns, eg, ceph/mon, gluster/glusterd etc. The global hierarchy should contain a mapping of roles to all the nodes that belong to that role. Each node itself simply lists the roles in the roles array.

When querying, this would enable the global listing to be queries for all the nodes that are running a ceph mon or glusterd etc. This can be cross referenced against a specific cluster's list of of nodes or against the node listing provided in an import cluster call to verify that all the nodes are being detected.

Also, the sds_pkg_name and sds_pkg_version don't belong. The detection should not be based on rpm packages. It should be based on processes, services, sockets or well known file locations, such as the cli. Versions must be queried by contacting the storage system directly, using the CLI or API call if known. In any case, the objects for the detected cluster are the proper place to put the package information if required, it doesn't belong in the node context.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+* The node agent component would introduce two python modules for ceph and
+gluster. These would be specific classes which extend from above SDSDetectPlugin
+and provide the actual implementation for the function
+
+```
+class DetectCephStorageSystem(SDSDetectPlugin):
+    def detect_storage_system(self):
+        # Implement the logic to get the details of underlying storage system
+        # 1. Figure out if ceph bits installed and set the type accordingly
```
brainfunked, line 166: As commented above, binaries, files, sockets, running processes etc. must be looked for, not just packages.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+        # 2. Figure the version of ceph bits installed and accordingly set the
```
brainfunked, line 167: Using the storage system's official facility to do so ie CLI or API. Package versions have no bearing on the actual version of the storage system and are not the de-facto source of this information. Storage system's interfaces must be used to determine this information.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+        # version
+        # 3. Figure out the role of the storage node (by figuring out if MON or
```
brainfunked, line 169: Wrong. One detection module detects only one specific role. Which means that there would be a separate module for ceph mon and ceph osd. Only the mon module would query for the cluster map to get the peer list, osd module would not. Each of these modules assigns it's own specific tag if the detection is successful.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+execute these plugins one by one to figure out the details of the underlying
+storage system. Tag the figured out details with the storage nodes in central
+store
+
+* SDS Detect plugin for `ceph`
+
+** Execute the command `ceph version` to figure out if ceph bits are installed
+and figure out the version details well
+
+** Check the running status of mon and osd processes on the storage nodes and
```
brainfunked, line 242: One specific type of process per module only.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+decide the role of the storage node accordingly
+
+** Get the FSID of the underlying cluster and tag the same as cluster_id. Store
```
brainfunked, line 245: This is the detected cluster id. Store it to clearly indicate as such. It is required only till a cluster is imported.

specs/sds_detection_modules.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+* SDS Detect plugin for `gluster`
+
+** Execute the command `gluster --version` to figure out if gluster bits are
+installed on the storage node. Also get the version details and tag with the
+node in central store
+
+** Default assign the role as `glusterd` for all the storage nodes
+
+** Execute the command `gluster pool list` and calculate a hash out of node
```
brainfunked, line 257: pool list?

brainfunked (changes requested):

specs/sds_detection_modules.adoc
shtripat: `gluster pool list` returns all the nodes in the cluster with uuids and was suggested by Atin.

specs/sds_detection_modules.adoc
shtripat: Will change the names a `detected_cluster_id` rather

specs/sds_detection_modules.adoc
shtripat: Done

specs/sds_detection_modules.adoc
shtripat: We would be using `ceph version` or `gluster --version` to get the details which is CLI. May be the wordings need to be changed a bit :)

specs/sds_detection_modules.adoc
japplewhite, line 257: Is there any negative checking here against a list of known supported configs: OS, Gluster and Ceph version, etc. Also the installation stories have lots of things we should check for like ntp, proper hostname setup, etc. Is this the appropriate place for that kind of thing?

Issue #126, opened 2017-01-30
Ceph CRUD Flows


brainfunked:

shtripat:
At the moment Calamri provides crud operations like create/edit/delete for storage pools only. For other entities like RBDs, erasure coded pools etc we need to consume CLI endpoint from Calamari.
The point I have is, do we really need Calamari (another daemon) running on MON nodes only for this purpose? Cant we look at using rados command execution directly from `ceph-integration` component itselt?

@Tendrl/tendrl-core suggestions/comments?

shtripat:
@GregMeno I have tried running Calamari on each monitor node (in a cluster with 3 mons and 3 osd nodes) and if I try to create a pool from one calamari session, I am able to see it reflected for other calamari properly.
Now the question I have is whether this model has some concerns or its OK. (I remember earlier intentionally we were running Calamari on only one mon node as they have own instances of DB running and there could be a mismatch between the data they are aware of). Is it fine if we use Calamari for posting CRUD operations to the ceph cluster with this model?

brainfunked:
@shtripat would you mind investigating the rados command availability for all the crud operations we're looking to implement and document them here?

shtripat:
The list of valid operations entity wise are -

1. Pool - delete/update/create
2. OSD - update/scrub/depp-scrub/update_config/repair
3. CRUSH - update
4. CRUSH Node - update/create/delete/add_bucket/remove_bucket/move_bucket/reweight_osd/move_osd

In addition to these we can run rados_command and rbd_command using utility available with https://github.com/Tendrl/ceph-integration/blob/master/tendrl/ceph_integration/ceph.py

@r0h4n correct if anything wrong or missed out here.

r0h4n:
@shtripat Yes, these are the capabilities http://docs.ceph.com/docs/giant/rados/api/python/

in addition to what you mentioned

shtripat:
Looks everything we need at the moment is achievable using rados and rbd command mechanism (if I am correct)

mbukatov:
> Looks everything we need at the moment is achievable using rados and rbd command mechanism (if I am correct)

That implies that Tendrl needs to be able to deploy ceph auth. keys (during cluster creation or import) for itself to be able to work with a ceph cluster. Do I understand this right?

ChristinaMeno:
@shtripat It's fine as long as you're not randomly doing each operation and pin requests and responses.
[...]

PR #134, opened 2017-02-07
Added spec ceph_crud_flows.adoc

shtripat:
tendrl-bug-id: Tendrl/specifications#126

### DIFF
--- a/specs/ceph_crud_flows.adoc
+++ b/specs/ceph_crud_flows.adoc
@@ -1,0 +1,557 @@
+// vim: tw=79
+
+= CRUD flows for ceph cluster entities
+
+The ceph-integration module need to provide flows for creating, retrieving,
+updating and deleting various entities like pools, rbds, ec-pools etc.
+
+== Problem description
+
+Currently the ceph-integration module provides flows only for creation and
+deletion of storage pools. Need to add flows for managing rbds, osd, erasure
+code profiles and ec-pools as well.
+
+== Use Cases
+
+CRUD operations for ceph cluster entities like storage pools, rbds, ec-pools and
+erasure code profiles.
+
+== Proposed change
+
+Add flows in ceph-integration module for below
+
+* Pool
+** Create
+** Update
+** Delete
+
+* RBDs
+** Create
+** Delete
+** Resize
+
+* Erasure coded pool
+** Create
+** Update
+** Delete
+
+* Erasure code profile
+** Create
+** Delete
+
+=== Alternatives
+
+There are two options to achieve the CRUD flows for ceph entities -
+
+* Use Calamari REST endpoints
+** pros
+*** Currently Calamari supports create/update/delete REST end points for pool
+management. For everything else, we need to invoke generic `CLI` end point from
+Calamari
+
+** cons
+*** Introduces another daemon on cluster MON nodes
+
+* Use direct rados and rnd command calls from ceph-integration modules
+** pros
+*** Doesn't need another daemon running on the MON nodes
+
+** cons
+*** None
+
+=== Data model impact:
+
+* Add models for
+** RBDs
+** Erasure coded pools
+** Erasure code profiles
+
+The snippet from definitions regarding the changes are as below
+
+```
+namespace.tendrl.ceph_integration:
+  flows:
+    .......
+    CreateECProfile:
+      atoms:
+        - tendrl.ceph_integration.objects.ECProfile.atoms.create.Create
+      help: "Create EC Profile"
+      enabled: true
+      inputs:
+        mandatory:
+          - ECProfile.name
+          - ECProfile.k
+          - ECProfile.m
+        optional:
+          - ECProfile.plugin
+          - ECProfile.directory
+          - ECProfile.ruleset_failure_domain
+      run: tendrl.ceph_integration.flows.create_ec_profile.CreateECProfile
+      type: Create
+      uuid: faeab231-69e9-4c9d-b5ef-a67ed057f98d
+    ......
+
+  objects:
+    Rbd:
+      atoms:
+        Create:
+          enabled: true
+          help: "Create rbd"
+          inputs:
+            mandatory:
+              - Rbd.pool_id
+              - Rbd.name
+              - Rbd.size
+          name: "Create Rbd"
+          run: tendrl.ceph_integration.objects.Rbd.atoms.create.Create
+          type: Create
+          uuid: 7a2df258-9b24-4fd3-a66f-ee346e2e3720
+        Delete:
+          enabled: true
+          help: "Delete rbd"
+          inputs:
+            mandatory:
+              - Rbd.pool_id
+              - Rbd.name
+          name: "Delete Rbd"
+          run: tendrl.ceph_integration.objects.Rbd.atoms.delete.Delete
+          type: Delete
+          uuid: 7a2df258-9b24-4fd3-a66f-ee346e2e3721
+        Resize:
+          enabled: true
+          help: "Resize Rbd"
+          inputs:
+            mandatory:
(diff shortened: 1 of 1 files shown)


shtripat:
@Tendrl/tendrl-core @Tendrl/tendrl_api please review

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+** Update
+** Delete
+
+* Erasure code profile
+** Create
+** Delete
+
+=== Alternatives
+
+There are two options to achieve the CRUD flows for ceph entities -
```
mbukatov, line 44: Another option - using RADOS python bindings - has been mentioned in https://github.com/Tendrl/specifications/issues/126#issuecomment-276754566

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+
+* Use Calamari REST endpoints
+** pros
+*** Currently Calamari supports create/update/delete REST end points for pool
+management. For everything else, we need to invoke generic `CLI` end point from
```
mbukatov, line 49: Or previously mentioned RADOS bindings.

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+      enabled: true
+      inputs:
+        mandatory:
+          - ECProfile.name
+          - ECProfile.k
+          - ECProfile.m
+        optional:
+          - ECProfile.plugin
+          - ECProfile.directory
+          - ECProfile.ruleset_failure_domain
```
mbukatov, line 88: If I read http://docs.ceph.com/docs/master/rados/operations/erasure-code-profile/ right (correct me if I'm wrong), one can specify any key value pair when creating EC profile:

> `{key=value}`
> Description:	The semantic of the remaining key/value pairs is defined by the erasure code plugin.

This is possible because each EC Profile Plugin has different set of keys. For example, see which plugins are supported by Red Hat for Ceph 2.0 - section "5.3. Erasure Code Plugins" in chapter [Erasure Code Pools ](https://access.redhat.com/documentation/en/red-hat-ceph-storage/2/paged/storage-strategies-guide/chapter-5-erasure-code-pools):

* Jerasure (Default) plugin has options `technique`, `ruleset-root`, `ruleset-failure-domain`
* Locally Repairable plugin has options: `l`, `ruleset-root`, `ruleset-locality`, `ruleset-failure-domain`
* ISA (Intel only) plugin: `technique`, `ruleset-root`, `ruleset-failure-domain`

Which means that actual list of inputs (and if those are mandatory or optional) are unique for each plugin, and not for EC profile in general.

Besides that, there is an option to define EC Profile using Locally Repairable plugin in low level way, which would require completely different model (see *Create a Low-level LRC Profile* section) - but this is probably beyond the original design presented here - I note this here to acknowledge this explicitly.

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+          enabled: true
+          help: "Create ec profile"
+          inputs:
+            mandatory:
+              - ECProfile.name
+              - ECProfile.k
+              - ECProfile.m
+            optional:
+              - ECProfile.plugin
+              - ECProfile.directory
```
mbukatov, line 212: The same issue with list of inputs (which is related to plugin used) as described above.

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+          type: int
+        plugin:
+          help: ec profile plugin
+          type: String
+        directory:
+          help: directory for ec profile
+          type: String
+        ruleset_failure_domain:
+          help: rule set failure domain for ec profile
+          type: String
```
mbukatov, line 257: The same issue with list of inputs (which is related to plugin used) as described above.

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+        format:
+          help: format of rbd
+          type: int
+        features:
+          help: featutres for rbd
+          type: list
+        flags:
+          help: flags for rbd
+          type: list
```
mbukatov, line 195: I see that there are more knobs available during rbd create:

> create (-s | –size size-in-M/G/T) [–image-format format-id] [–object-size size-in-B/K/M] [–stripe-unit size-in-B/K/M –stripe-count num] [–image-feature feature-name]... [–image-shared] image-spec

From http://docs.ceph.com/docs/master/man/8/rbd/

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
+// vim: tw=79
+
+= CRUD flows for ceph cluster entities
+
+The ceph-integration module need to provide flows for creating, retrieving,
+updating and deleting various entities like pools, rbds, ec-pools etc.
```
mbukatov, line 6: Do we plan to address snapshots in this specification as well?

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+          type: String
+        min_size:
+          help: "sets the number of replicas for objects in the pool"
+          type: Integer
+        pg_num:
+          help: "The total number of placement groups for placement purposes."
+          type: Integer
+        pgp_num:
+          help: "The total number of placement groups for the pool."
+          type: Integer
```
mbukatov, line 357: pgp-num attribute is listed, but it's not included in any flow/atom - is it ok?

specs/ceph_crud_flows.adoc
```diff
@@ -0,0 +1,557 @@
...
+              - Pool.poolname
+              - Pool.size
+              - Pool.pg_num
+              - Pool.quota_enabled
+              - Pool.quota_max_objects
+              - Pool.quota_max_bytes
+          run: tendrl.ceph_integration.objects.Pool.flows.update_pool.UpdatePool
+          type: Update
+          uuid: 4ac41d8f-a0cf-420a-b2fe-18761e07f3b2
+      attrs:
```
mbukatov, line 342: I don't see eg. expected-num-object listed there. We may need to check how is this aligned with http://docs.ceph.com/docs/master/rados/operations/pools/

mbukatov (commented):
@mkudlej @shtripat I went through the specification and added notes when something attracted my attention. This is not a full review.

specs/ceph_crud_flows.adoc
shtripat: Snapshots is not part of this spec. Only pool/rbds/ec-profiles are taken care as part of this spec.

specs/ceph_crud_flows.adoc
shtripat: Ok. Evaluated the two mentioned here at the moment.

specs/ceph_crud_flows.adoc
shtripat: Correct

specs/ceph_crud_flows.adoc
shtripat: The atoms are created for the time being but not exposed from tendrl UI. The default ec-profiles with 4+2, 6+3 and 8+4 would be created.

specs/ceph_crud_flows.adoc
shtripat: Of course there are but for the time being just supporting `size` only as exposed from tendrl.

specs/ceph_crud_flows.adoc
shtripat: pgp_num is actually set the same value that of pg_num. Its present here only for listing purpose.

PR #135, opened 2017-02-08
spec for create RBD

gnehapk:


specs/create-rbd-spec.adoc
```diff
@@ -0,0 +1,236 @@
...
+* If the target size * RBDs to create > available storage in this cluster, a warning message will be displayed, but 
+  the user will be allowed to continue.
+
+* For block devices, a backing pool must be created. Only one pool can be created per set of block devices.
+
+* In the second step, user will decide whether to use an existing pool or create a new pool as the backing pool
+  for this set of RBD.
+  (https://redhat.invisionapp.com/share/EHAC6S0N4#/screens/217735638)
+
+* On choosing existing pool, a drop down will be displayed which will list down all the available pools in the
```
gnehapk, line 50: @anivargi Please confirm here i.e. the drop down will display all the available pools in the Tendrl.

specs/create-rbd-spec.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+    - OSDs
+
+    - Replicas
+
+    - PG Count
+
+    - Journal Configuration
+
+    - Quotas
```
gnehapk, line 71: @anivargi type, OSDs, replicas, journal configuration and quotas properties of pool are missing in the API response as per your documentation (https://github.com/Tendrl/api/blob/d3b44bbd8f85f545e112f0fae781fed3b366d3c2/docs/clusters.adoc#list-clusters).

specs/create-rbd-spec.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+== Proposed change
+
+Front-end related changes:
+
+    * Create view for "Create RBD" workflow.
+    * An API call be triggered at the bootstrap of application which will fetch the cluster list, pool list 
+      to display the 
+
+      - cluster options and size for each cluster in step one
```
gnehapk, line 117: @anivargi Cluster's size is missing in API response. Please provide us the updated API response.

specs/create-rbd-spec.adoc
```diff
@@ -0,0 +1,236 @@
...
+                    "used": "0",
+                    "min_size": "2"
+                }
+            }
+        }]
+}
+
+
+API to post the data for import cluster - Required.
+
```
gnehapk, line 166: @anivargi Please provide the POST API call for creating RBD.

gnehapk:
@anivargi @brainfunked @nthomas-redhat @shtripat @kamleshverma1 @cloudbehl Please review.

specs/create-rbd-spec.adoc
```diff
@@ -0,0 +1,236 @@
...
+    - PG Count
+
+    - Journal Configuration
+
+    - Quotas
+
+* If user wants to create a new pool, they are presented with a pool creation form. See Create Pool Workflow's spec 
+  for details about these options.
+  (https://redhat.invisionapp.com/share/EHAC6S0N4#/screens/217735639)
```
gnehapk, line 75: @anivargi So first the UI needs to send the POST call for creating pool in the 2nd step and after that only POST call for creating RBD will be sent in the last step. Please confirm whether my assumption is correct or not.

anivargi:
@gnehapk All the docs are in this PR https://github.com/Tendrl/api/pull/76

gnehapk:
@anivargi Thanks. Please answer other questions too which are not for response format.

specs/create-rbd-spec.adoc
shtripat: RBDs can be created on replicated pool only I feel. We should list only replicated type pools here.

specs/create-rbd-spec.adoc
shtripat: @gnehapk @anivargi For this I feel we need pool utilization data to be attached to cluster and `Utilization.Available` need to be shown here. Will create a PR for the back-end changes as its not available at the moment. Only pool utilization data is available for individual pools.

shtripat: Added https://github.com/Tendrl/ceph-integration/issues/110 for the same.

specs/create-rbd-spec.adoc
julienlim: It is not possible to create an RBD image on an erasure coded pool because it requires partial writes. It is however possible to create an RBD image on an erasure coded pools when a replicated pool tier set a cache tier.  See http://docs.ceph.com/docs/master/rados/operations/erasure-code/.  Agree that only replicated pools get listed.

specs/create-rbd-spec.adoc
julienlim: You will also need pool flag values to know what's set or not to determine if certain Day 2 operations are permitted.

PR #144, opened 2017-02-27
API test coverage

mkudlej:
I've tried to describe in this specification need for comparing couple of different sources for API function lists.

### DIFF
--- a/specs/test_api_function_coverage.adoc
+++ b/specs/test_api_function_coverage.adoc
@@ -2,7 +2,7 @@
 
 
 This specification goal is to describe changes needed for implement
-tracking of API functions test coverage.
+tracking of the API functions test coverage.
 
 
 == Problem description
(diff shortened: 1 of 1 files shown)


mkudlej:
@brainfunked @anivargi Please review.

r0h4n:
@mkudlej whats the status of this one, seems stale

fbalak:
We have parsing scripts but they are deprecated now and we need to update them with new paths for definition files and our testing framework. There are now tasks with higher priority that we are working on. We will revisit this item after they will be done.

r0h4n:
Any updates on this?

fbalak:
@r0h4n We have some [script](https://github.com/fbalak/apistate/blob/master/apistate.py) used in downstream. It generates table from [usmqe-tests](https://github.com/Tendrl/usmqe-setup) and [api](https://github.com/Tendrl/api) repos. It lists api calls covered in our framework, in our tests and documented calls.

r0h4n:
I am going to close this one, if required open another spec with specific dates and targets

PR #146, opened 2017-03-08
[WIP] Ceph create cluster parameters and validations

brainfunked:


specs/ceph_installer_wrapper.adoc
```diff
@@ -54,8 +54,45 @@ Tendrl API provides an interface to create a Ceph cluster.
...
-
+POST '/api/1.0/CreateCluster'
+
+{
+  "sds_type": "ceph",
+  "sds_version": "10.2.5",
+  "name": "MyCluster",
+  "node_configuration": {
+    "3a95fd96-876d-439a-a64d-70332c069aaa": {
+      "role": "ceph/mon"
```
shtripat: For mon node, configure call needs details like `fsid`, `name`, `cluster_network`, `public_network`, `interface`. I feel these fields also should be passed here while create cluster flow either as part of top level fields or mon-node details.

Generated `fsid` for cluster could be top level, whereas `interface`, `cluster_network` and `public_network` are specific to node, so should be passed as part of mon-node. The generated `fsid` should be written at `/etc/tendrl/{integration name}/integration_id` once cluster creation is successful.

specs/ceph_installer_wrapper.adoc
```diff
@@ -54,8 +54,45 @@ Tendrl API provides an interface to create a Ceph cluster.
...
+    },
+    "3943fab1-9ed2-4eb6-8121-5a69499c4568": {
```
shtripat: osd node details should contain `public_network` and `cluster_network` values as well. These would be provided while create-cluster flow from UI and should be passed down here through api layer. Same applies for mon node details as well.

specs/ceph_installer_wrapper.adoc
```diff
@@ -92,48 +103,78 @@ Status: 202 Accepted
...
+    "sds_type": "ceph",
+    "sds_version": "10.2.5",
+    "name": "MyCluster",
+    "node_ids": [],
+    "tags": ["provisioner/ceph"],
+    "parameters": {
+      "TendrlContext.integration_id": "9a4b84e0-17b3-4543-af9f-e42000c52bfc",
+      "Node[]": ["3943fab1-9ed2-4eb6-8121-5a69499c4568"],
+      "node_configuration": {
+        "3943fab1-9ed2-4eb6-8121-5a69499c4568": {
```
shtripat: These details need update as mentioned in above comments.

shtripat: For required parameters plz refer https://github.com/Tendrl/node-agent/pull/232

mbukatov:
I have few question not directly related to the changes proposed in this pull request, but since the spec document is called **Tendrl Ceph-installer Wrapper** I would expect to see answers to the following questions here. If I'm wrong about that and this is not right place for such details, feel free to redirect me somewhere else:

* How would tendrl user get to the ansible logs from ceph-installer? By that I mean stdout from ansible. Something like that is extremely useful for debugging. Would that require Tendrl API call?
* Would it be possible to force stop the task? Eg. when ansible get stuck because of some error?
* What would Tendrl do when ceph-installer crashes, fails or misbehaves in so other way?
* Would it be possible to restart ceph-installer task later? Eg. I have a lab with 20 servers and start the task, but some were not provisioned correctly and ansible fails there. So I go and fix the problem on affected servers and now what?
* How would you handle parallel execution so that when Tendrl runs ceph-ansible task, it will be executed on all storage machines at the same time? How would that affect design of flows and data model?

mbukatov:
Yeah, and one more thing: how would crush cluster hierarchy affect this?

specs/ceph_installer_wrapper.adoc
```diff
@@ -37,103 +37,330 @@ A wrapper needs to be created which abstracts the capabilities provided by the c
...
+    }
+  }
+}
+----
+
+The validations to be carried out on these parameters are as follows, when the
+role of a node is `ceph/osd`:
+
+* `journal_size` is optional. Value should be in Megabytes.
+* `journal_colocation` is optional and value should be a boolean.
```
mbukatov: Is `"journal_colocation": "true"` and `"journal": "/dev/sdc"` valid configuration?

specs/ceph_installer_wrapper.adoc
```diff
@@ -37,103 +37,330 @@ A wrapper needs to be created which abstracts the capabilities provided by the c
...
+** It has to contain hashes.
+** Each hash has to contain a `device` key.
+*** If `journal_colocation` is set to `true` _for the node_, the `journal` key
+    cannot exist in the hash
+*** If `journal_colocation` is set to `false` _for the node_, the `journal` key
+    must contain a device
+** Every device path specified in either the `device` or the `journal` key must
+   be a valid block device on the node
+** A device cannot be listed under both `device` and `journal` (across multiple
+   hashes), per host
```
mbukatov, line 477: Single device should not be listed under different hashes, eg. this seems wrong to me and it would be easy to verify:

```
"storage_disks": [
  {
    "device": "/dev/sda"
  },
  {
    "device": "/dev/sda"
  }
```

specs/ceph_installer_wrapper.adoc
```diff
@@ -37,103 +37,330 @@ A wrapper needs to be created which abstracts the capabilities provided by the c
...
+          }
+        ]
+      }
+    }
+  }
+}
+----
+
+The validations to be carried out on these parameters are as follows, when the
+role of a node is `ceph/osd`:
```
mbukatov: Additional validation ideas include:

* total size on journal dist requested (based on `journal_size` and number of OSDs sharing given journal device) vs total size of that device

That said, I have no idea where this validation would be performed so I don't know if you will know this already during validation.

specs/ceph_installer_wrapper.adoc
brainfunked: Ack, updating the spec for this.

UPDATE: Spec has been updated to include this.

r0h4n:
@brainfunked Is this still required?

PR #147, opened 2017-03-14
spec for create file share

gnehapk:


specs/create-file-share-spec.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+    * API to create file share - Required
+
+    * API to get list of bricks - Required
+
+    * API to create brick - Required
```
gnehapk, line 39: @anivargi Please provide the mentioned APIs.

gnehapk:
@anivargi @shtripat @nthomas-redhat @kamleshverma1 @cloudbehl  @nnDarshan  Please review. I have mentioned only the things that we need from API in UI. Will detail this spec later on.

@nnDarshan @shtripat  Which access protocols are supported? Do we have cache options and Thread options(https://redhat.invisionapp.com/share/Q78YMAVDJ#/screens/197076159) data available which needs to be shown on the UI

@anivargi Will it be possible to implement Users and permission in API as we don’t have support for user management in Tendrl? It is required in UX design- https://redhat.invisionapp.com/share/Q78YMAVDJ#/screens/197076156

nnDarshan:
@gnehapk gluster is the access protocol thats supported for now.

Regarding cache options that is not available for now...even the UX design page mentions that need not be part of mvp. So we can add that later

specs/create-file-share-spec.adoc
gnehapk: @anivargi Please provide the API for getting list of disks as well.

r0h4n:
@gnehapk Is this done?

PR #152, opened 2017-04-07
WIP - Added spec ceph_journal_mapping.adoc

shtripat:
tendrl-bug-id: Tendrl/specifications#151

### DIFF
--- a/specs/ceph_journal_mapping.adoc
+++ b/specs/ceph_journal_mapping.adoc
@@ -1,0 +1,371 @@
+// vim: tw=79
+
+= Flow for generating journal mapping while create cluster workflow in Ceph
+
+A flow needs to added in `ceph-installer` provisioner of `tendrl-node-agent`
+which generates a possible journal mapping based on standard guidelines from
+Ceph. This mapping would be just a suggestion, and user can always change the
+mapping and pass on to create cluster flow.
+
+This helps end user to visualize how OSDs and their journals would be layout
+in the final cluster getting created. Using this visualization user can always
+change the maping based on his/her needs and submit for cluster creation flow.
+
+== Problem description
+
+While create cluster flow in case of Ceph user needs to visualize and form a
+mapping of OSDs and journals in case of dedicated journals and provide the
+details to create cluster flow.
+
+If Tendrl provides a suggestion based on Ceph standard guidelines it would be
+easier for end user to visualize how the journal mapping would look in final
+cluster. Also if user doesn't agree to the mapping, he can always change the
+same before submitting for cluster creation.
+
+== Use Cases
+
+Provide a visualization of journal mapping for OSDs while create cluster flow of
+Ceph.
+
+== Proposed change
+
+Below changes should be done to provide the above mentioned functionality in
+Tendrl -
+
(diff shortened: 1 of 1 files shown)


shtripat:
@Tendrl/tendrl-core plz review

specs/ceph_journal_mapping.adoc
```diff
@@ -0,0 +1,371 @@
...
+        "node_id_2": {
+            "storage_disks": [
+                {"device": "/dev/vdb", "size": bytes, "ssd": true },
+                {"device": "/dev/vdc", "size": bytes, "ssd": false},
+                ...
+            ]
+        },
+        .....
+    }
+}
```
gnehapk, line 61: @shtripat @anivargi "node_configuration" contain the list of nodes with their configuration. So, IMO "node_configuration" should be an array instead of nested objects.

specs/ceph_journal_mapping.adoc
```diff
@@ -0,0 +1,371 @@
...
+        "node_id_2": {
+            "storage_disks": [
+                {"device": "/dev/vdb", "journal": "/dev/vdc" },
+                {"device": "/dev/vdd", "journal": "/dev/vde"},
+                ...
+            ],
+            "unallocated_disks": ["/dev/vdf", "/dev/vdg"]
+         },
+         ...
+}
```
gnehapk, line 85: @shtripat @anivargi same as the above comments. In case of array, UI can directly iterate over its length else we have to get the keys length and on it the UI needs to iterate.

specs/ceph_journal_mapping.adoc
shtripat: For me array/dict both are a collection of nodes. Its just how you represent them.
With everybody feel it should list of dicts, can be changed. Earlier it was just kept inline with create cluster flow so its like this at the moment.

specs/ceph_journal_mapping.adoc
gnehapk: @shtripat @anivargi Is it possible to send the node_id value against some property instead of the key name as this becomes difficult in the UI to read. 
If the node_id value is coming against some property, UI can directly take that value and update the mentioned node with the journal details.  Else we need to first iterate over the key names to find the node_id and then to update the journal details

Sample format -
```
{
      journal_details: [{
              node_id: "node_id_1",
             "storage_disks": [
                 {"device": "/dev/vdb", "journal": "/dev/vdc" },
                 {"device": "/dev/vdd", "journal": "/dev/vde"},
                 ...
             ],
             "unallocated_disks": ["/dev/vdf", "/dev/vdg"]
         }, {
             node_id: "node_id_2",
             "storage_disks": [
                 {"device": "/dev/vdb", "journal": "/dev/vdc" },
                 {"device": "/dev/vdd", "journal": "/dev/vde"},
                 ...
             ],
             "unallocated_disks": ["/dev/vdf", "/dev/vdg"]
          }]
 }
```

@anivargi Is it possible for API to send the response like this to UI?

r0h4n:
@cloudbehl @shtripat @gnehapk The backend changes have already been merged. What else remains on this one? Or are we done with this spec?

PR #159, opened 2017-05-16
Spec disk inventory etcd structure

GowthamShanmugam:


GowthamShanmugam:
@brainfunked @r0h4n @nnDarshan please review

GowthamShanmugam:
@brainfunked @r0h4n @shtripat @nnDarshan hwinfo gives
Device Files: /dev/vdh, /dev/disk/by-id/virtio-78065dbf-92db-4574-8, /dev/disk/by-path/virtio-pci-0000:00:0e.0
here /dev/disk/by-id/virtio-78065dbf-92db-4574-8 is never change for that particular disk even disk name change when we do deactivate disk.
we can use this, but problem is for real machine it gives two id
like
Device Files: /dev/sda, /dev/disk/by-id/ata-SAMSUNG_MZ7TE512HMHP-000L1_S1GJNSAG400778, /dev/disk/by-id/wwn-0x4d30445853885002

both format looks different

GowthamShanmugam:
Here only problem is id

GowthamShanmugam:
Some unique id is needed we can't use (manufacturer-id+vendor-id+device-serial-number), Because Vm does not have vendor details

shtripat:
@GowthamShanmugam I feel `/dev/disk/by-id/ata-SAMSUNG_MZ7TE512HMHP-000L1_S1GJNSAG400778` is more meaningful if you can get and use.

GowthamShanmugam:
@shtripat in VM it comes like  /dev/disk/by-id/virtio-78065dbf-92db-4574-8 is that ok?

shtripat:
@GowthamShanmugam I feel that should be fine as long as it doesnt change on removal and addition of new disks in between the list.

@Tendrl/tendrl-core comments?

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -1,121 +1,319 @@
...
+            phy_sector_size: "512"
+            log_sector_size: "512"
+            ssd: "True"
+            scheduler_name: "cfq"
+            req_queue_size: "128"
+            discard_align_offset: "0"
+            discard_granularity: "0"
+	    discard_max_bytes: "0"
+	    discard_zeros_data: "0"
+	    partition:{
```
brainfunked, line 264: partitions (plural) needs to be an array of dictionaries, one per partition.

specs/disk_details_for_hardware_inventory.adoc
```diff
@@ -1,121 +1,319 @@
...
+            		discard_max_bytes: "0"
+            		discard_zeros_data: "0"
+		}
+
+	}
+	"disk_id2": {
+		//same as disk_id1 structure
+	}
+   }
+   "all_block_device":{
```
brainfunked, line 304: Where exactly do these block device related keys get stored?

brainfunked (changes requested):
Document the sync strategy for each of the pieces.

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked i will change it

specs/disk_details_for_hardware_inventory.adoc
GowthamShanmugam: @brainfunked it will stored under 
 all_block_device - nodes/$Node_context.node_id/block_devices/all
 used_block_device - nodes/$Node_context.node_id/block_devices/used
  free_block_device  - nodes/$Node_context.node_id/block_devices/free

brainfunked:
The disk id for a physical disk is assembled using the "Vendor", "Device" and "Serial ID" attributes.

If these three fields are absent, check if the values of the "Driver" attribute are `virtio*`. If yes, then use the value of the virtio-* part from the "Device Files" list where the first value in the list matches the regex `.+\/by-id\/(virtio-.+)`. The match 1 would contain the string virtio-* which would be used as the id.

GowthamShanmugam:
@brainfunked @r0h4n @nnDarshan if we store disk(hwinfo), block_device(lsblk) then free disk will not have disk_size and ssd or not. Because hwinfo is not providing these details. 

One more issue is we can use block_device object to append lsblk details with disk because, key for block device should be one per object but this case key comes twice like it has to store under disk and partition also.

Is hashing will solve the issue like disk directory contain  two disk details but locally collected  disk detail have one disk detail, is that extra one is deleted from etcd. what kind of approach for that hashing. because  i have to delete extra disk under /Disks and extra partition details for particular Disk.

/dev/disk gives   by-id/    by-uuid/  for some machine      by-id/        by-partlabel/    by-partuuid/     by-path/      by-uuid/
so it is not fixed am i create object for this like
class RawReferences contains attr:
      reference (json)
       Here json structure like 
                     {
                       "by-id": "all by id details as single string",
                       "by-uuid": "all uuid details as single string"
                         .
                         .
                         .
                    }

Issue #167, opened 2017-07-17
Enhance gluster cluster sync logic


shtripat:
We need to enhance/modify the gluster cluster sync logic for
- Capture volume snapshot details and update in central store
- Instead of using `gstatus` aggregate the peer and volume status to decide the cluster's healthy status
- Instead of using `gstatus` aggregate the brick utilization for figuring out the volume utilization details
- Use `statvfs` for brick mounts to figure out the brick utilization - DONE
- Identify the arbiter volume type based on volme{n}.arbiter_count value
- Add rebalance estimation time under rebalance details
- No of geo-rep sessions and active geo-rep sessions
- Client connections per brick and aggregate for a volume (check if connections per brick available in get-state output)

shtripat:
@Tendrl/tendrl-core plz review and add if something missed out here.

nnDarshan:
@shtripat can you please add geo-rep

shtripat:
@nnDarshan also I think the way volume options would be reported would undergo changes in get-state.
Also client connections details to be added.

shtripat:
Closing this as all the required changes are merged including specifications PR as well.

PR #170, opened 2017-07-18
Added specification enhance_gluster_cluster_sync_logic.adoc

shtripat:
tendrl-bug-id: Tendrl/specifications#167

### DIFF
--- a/specs/enhance_gluster_cluster_sync_logic.adoc
+++ b/specs/enhance_gluster_cluster_sync_logic.adoc
@@ -1,0 +1,531 @@
+= Enhance gluster cluster synchronization logic as per recommendations
+
+The intent of this change is to accommodate additional details available as part
+of `get-state` output while syncing the gluster cluster details and store them
+in central store.
+
+Also instead of using `gstatus` for cluster health and volume utilization
+details, we need to user recommendations provided by gluster team.
+
+
+== Problem description
+
+This specification talks about various changes needed to be done under gluster
+cluster synchronization code to accomodate the additional details related to
+volume snapshots, geo-rep sessions.
+
+Also as `gtstaus` is quite a unstable and also few scenarios are not taken care
+of, using it for deciding the overall cluster status and volume utilization is
+not appropriate. Rather we need to write the aggregations logic to decide the
+cluster state based on peers status and volume status of the gluster trusted
+storage pool. Also the overall volume utilization should be calculated based on
+aggregations done for the participating bricks of the volumes.
+
+
+== Use Cases
+
+This addresses the synchronization of gluster cluster details like
+
+* Volume snapshot details
+* Add volume re-balance estimation time details
+* Volume geo-rep details
+* Client connections per volume
+* Identification of arbiter volumes in cluster (if any)
+* Instead of `gstatus` use aggregations across bricks utilization to calculate
+volume utilization
+* Instead of `gstatus` use peer status and volume status to aggregate and
+calculate the cluster status
+
+
+== Proposed change
+
+* Use `get-state detail` and `get-state volumeoptions` commands instead of plain
+`get-state` as first provides the detailed (e.g. client connections etc.)
+information for the volume and second is option to list all the volume options
+with default values
+
+* Use `get-state` output to figure out the volume snapshot details and update in
+central store. The snapshot details available as part of `get-state` output are
+documented under references section in sample. These details should be used to
+create instances of `Snapshot` class and persist for a volume.
+
+* Calculate the estimation time for re-balance job. This should be largest
+value among the `time_left` values reported by get-state from different nodes of
+the cluster
+
+* Use `get-state` output for geo-rep session details like no. of geo-rep
+sessions and active geo-rep sessions and update per volume in central store
+
+* Figure out the client connections per brick from `get-state` output and
+aggregate for volume and update in central store. The no of client connections
+for a volume merely a sum of all the participating brick's client connections.
+Note: This logic for calculating client connections for a volume is trivial but
+its as per the guidelines provided from Gluster team.
+
+* Identify an arbiter volume and mark accordingly in central store, based on
+field `volume{n}.arbiter_count` in `get-state` output. For a volume if value of
+`volume{n}.arbiter_count` is non-zero, the volume should be marked as arbiter
+type
+
+* Remove the uses of `gstatus` and use aggregations across bricks of the volume
+to decide the volume utilization data.
+
+The logic here goes like
+
+** First calculate the `raw_capacity` and `raw_used` values for volume and
+cluster merely by adding the `raw_capacity` and `raw_used` for bricks of the
+volume. These details are synced and available with central store for the
+individual bricks
+
+** Now the pseudo code to decide the final values of volume and cluster usable
+and used capacity goes as below
+
+```
+if no of up bricks == no of total bricks:
+     if  volume's disperse_count == 0:
+         # This is distribute or replicate volume
+         volume's usable_capacity = volume's raw_capacity / volume's replica_count
+         volume's used_capacity = volume's raw_used / volume's replica_count
+     else:
+         # this is a disperse volume, with all bricks on-line
+         # assumption : all bricks are the same size
+         disperse_yield = \
+            (volume's disperse_count - volume's redundancy_count)
+            / volume's disperse_count
+         volume's usable_capacity = volume's raw_capacity * disperse_yield
+         volume's used_capacity = volume's raw_used * disperse_yield
+ else:
+     if volume's replica_count > 1 or volume's disperse_count > 0:
+         for subvol in volume's subvols:
+             for brick in subvol:
+                 if brick.status == "Started":
+                     volume's usable_capacity += brick's utilization['total']
+                     volume's used_capacity += brick's utilization['used']
+                 # For replicate volume use only one replica
+                 if volume's replica_count > 1:
+                     break
+     else:
+         volume's usable_capacity = volume's raw_capacity
+         volume's used_capacity = volume' raw_used
+ if volume's usable_capacity > 0:
+     volume's pcnt_used = (volume's used_capacity / volume's usable_capacity) * 100
+ else:
+     volume's pcnt_used = 0
+
+ cluster's usable_capacity += volume's usable_capacity
+ cluster's pcnt_used = 0
+ if cluster's usable_capacity > 0:
+     cluster's pcnt_used = (cluster's used_capacity / cluster's usable_capacity) * 100
+```
+
+Note: This logic is merely lifted from gstatus as suggested by Gluster team
+
+* Remove the uses of `gstatus` and use aggregations of `gluster peer status` and
+`gluster volume status` to decide the cluster health status
+
+The here goes like
+
+** First calculate the volume states based on volume's individual bricks status
+values. The pseudo code to explain the state calculation for volumes is as below
+
+```
+Form a map of sub-volumes with their no of down bricks
+
+if no of up bricks in volume is 0:
+    volume's state = 'down'
+else:
+    if volume's replica_count > 1 or volume's disperse_count > 0:
+        find the subvolume with most down bricks as worst subvolume
+        if no of down bricks in worst subvolume > 0:
+            subvolume problem count = max(
+                volume's replica count,
+                volume's redundancy count + 1
+            )
+            if no of down bricks in worst subvolume == subvolume problem count:
+                if only one subvolume in volume:
+                    volume's state = 'down'
+                else:
+                    volume's state = 'partial'
+            else:
+                volume's state = 'degraded'
+    else:
+        # this volume is not protected (replicate/disperse) so any brick down
+        # means volume state partial
+        if no of up bricks != total bricks:
+            volume's state = 'partial'
+```
+
+Note: This logic is merely lifted from gstatus as suggested by Gluster team
+
+** If any of the volume's state is down or partially up, the cluster status is
+marked as `unhealthy`
+
+** After above logic applied, the gluster peer's collected state is considered
+to overwrite the cluster status. If any of the peer is in non-connected state,
+the cluster status is marked as `unhealthy`. This is to be done un-conditionally
+because in case where there are no volumes in the cluster yet, the peer's
+connected state is the deciding factor for the cluster status
+
+Note: This logic us merely lifted from gstatus as suggested by Gluster team
+
+* Add additonal field `state` under `Volume` object. Use aggregated bricks
+status to find out the volume state value. The logic above defines the mechanism
+how the state of the volume to be decided. Valid set of states for a volume
+would be
+
+** up
+** down
+** up(partial)
+** up(degraded)
+** unknown
+
+Note: The value `unknown` is applicable for the volumes till all the bricks
+details are available in central store.
+
+=== Alternatives
+
+None
+
+=== Data model impact
+
+* Add an additional object `Snapshot` to hold volume snapshots details
+
+```
+Snapshot:
+  attrs:
+    vol_id:
+      help: Volume id
+      type: String
+    name:
+      help: name of the snapshot
+      type: String
+    id:
+      help: uuid of the snapshot
+      type: String
+    time:
+      help: Creation time of the snapshot
+      type: String
+    description:
+      help: description of snapshot
+      type: String
+    status:
+      help: status of snapshot
+      type: String
+  enabled: true
+  value: clusters/$TendrlContext.integration_id/Volumes/$Volume.vol_id/Snapshots/$Snapshot.id
+  list: clusters/$TendrlContext.integration_id/Volumes/$Volume.vol_id/Snapshots
+help: gluster volume snapshot details
+```
+
+* Add additional object `ClientConection` to hold the brick connections details
+
+```
+ClientConnection:
+  attrs:
+    brick_name:
+      help: name of the brick
+      type: String
+    hostname:
+      help: client hostname
+      type: String
+    bytesread:
+      help: no of bytes read
+      type: int
+    byteswrite:
+      help: no of bytes written
+      type: int
+    opversion:
+      help: operational version
+      type: String
+  enabled: true
+  value: clusters/$TendrlContext.integration_id/Bricks/all/$Brick.brick_name/ClientConnections/$ClientConnection.hostname
+  list:  clusters/$TendrlContext.integration_id/Bricks/all/$Brick.brick_name/ClientConnections
+  help: brick client connection details
+```
+
+* Add additional field `client_count` under `Volume` and `Brick` classes to hold
+client connection count values
+
+* Add an additional field `time_left` under `RebalanceDetails` object to capture
+remaining time for rebalance
+
+* Add additional field `rebal_estimated_time` under `Volume` object to capture
+the estimated time to complete rebalance on the volume
+
+* Add additional field `state` under `Volume` object to hold the derived volume
+state based on participating brick's status values
+
+* Add additional field `is_arbiter` under `Brick` object to mark an arbiter
+brick
+
+* Add an additional object `GeoReplicationPair` to hold the geo-rep session pair
+details for gluster volumes
+
+```
+GeoReplicationPair:
+    attrs:
+        vol_id:
+          help: Volume id
+          type: String
+        session_id:
+          help: "unique id of geo replication session"
+          type: String
+        pair:
+          help: "geo replication session pair name"
+          type: String
+        master_node:
+          help: "master node ip/fqdn"
+          typoe: String
+        master_volume:
+          help: "master volume name"
+          type: String
+        master_brick:
+          help: "master brick name"
+          type: String
+        slave_user:
+          help: "slave user name"
+          type: String
+        slave:
+          help: "slave host and volume name"
+          type: String
+        slave_node:
+          help: "slave node with which geo-rep session is going"
+          type: String
+        status:
+          help: "geo replication session status"
+          type: String
+        crawl_status:
+          help: "geo replication crawl status"
+          typoe: String
+        last_synced:
+          help: "last synced time"
+          type: String
+        entry:
+          help: "entries synced"
+          type: String
+        data:
+          help: "data synced"
+          type: String
+        meta:
+          help: "metadata synced"
+          type: String
+        failures:
+          help: "number of failures"
+          type: String
+        checkpoint_time:
+          help: "checkpoint time"
+          type: String
+        checkpoint_completed:
+          help: "if checkpoint is completed"
+          type: String
+        checkpoint_completion_time:
+          help: "time of checkpoint completion"
+          type: String
+      enabled: true
+      value: clusters/$TendrlContext.integration_id/Volumes/$Volume.vol_id/GeoReplicationSessions/$GeoReplicationPair.session_id/$GeoReplicationPair.pair
+      list: clusters/$TendrlContext.integration_id/Volumes/$Volume.vol_id/GeoReplicationSessions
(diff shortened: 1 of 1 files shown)


shtripat:
@Tendrl/tendrl-core @atinmu @samikshan plz review.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,275 @@
...
+* Figure out the re-balance estimation time from `get-state` output and update
+for volume in central store
+
+* Use `get-state` output for geo-rep session details like no. of geo-rep
+sessions and active geo-rep sessions and update per volume in central store
+
+* Figure out the client connections per brick from `get-state` output and
+aggregate for volume and update in central store
+
+* Identify a arbiter volume and mark accordingly in central store, based on
```
nnDarshan: We might have to mark the brick that takes the role of arbiter as well. This would involve modification to brick object. Can you please add some details around this.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,275 @@
...
+* Identification of arbiter volumes in cluster (if any)
+* Instead of `gstatus` use aggregations across bricks utilization to calculate
+volume utilization
+* Instead of `gstatus` use peer status and volume status to aggregate and
+calculate the cluster status
+
+
+== Proposed change
+
+* Use `get-state` output to figure out the volume snapshot details and update in
```
nnDarshan: Since some snapshot details like snap hard/soft limit are available as part of `gluster get-state --volume-options` (I am guessing the options would be available by some command like this). How are we planning to club the details from these 2 commands. Since the above mentioned command is something that we should invoke only once. Can you please add some details around the strategy to collect those details.

shtripat: I feel snapshot quota options are volume options and they can still be maintained under volume options entity and not as part of volume snapshots. Volume snapshots entity maintains only the list of snapshot (if any) for the volume. Hope this makes sense?

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Sure. We can have a flag `arbiter_brick` set as boolean at brick level. Will add that.

shtripat: In `get-state` output we do get the `arbiter_count = 1` for an arbiter volume but there is no marking of the arbiter brick as such. A sample output is as below

```
[Volumes]
Volume1.name: arb-vol
Volume1.id: 543fd98b-b876-4027-826a-c051761b5b40
Volume1.type: Replicate
Volume1.transport_type: tcp
Volume1.status: Started
Volume1.brickcount: 3
Volume1.Brick1.path: dhcp43-118.lab.eng.blr.redhat.com:/root/gluster_bricks/vol3_b1
Volume1.Brick1.hostname: dhcp43-118.lab.eng.blr.redhat.com
Volume1.Brick2.path: dhcp43-116.lab.eng.blr.redhat.com:/root/gluster_bricks/vol3_b2
Volume1.Brick2.hostname: dhcp43-116.lab.eng.blr.redhat.com
Volume1.Brick3.path: dhcp43-126.lab.eng.blr.redhat.com:/root/gluster_bricks/vol3_b3
Volume1.Brick3.hostname: dhcp43-126.lab.eng.blr.redhat.com
Volume1.Brick3.port: 49152
Volume1.Brick3.rdma_port: 0
Volume1.Brick3.status: Started
Volume1.Brick3.signedin: True
Volume1.snap_count: 0
Volume1.stripe_count: 1
Volume1.replica_count: 3
Volume1.subvol_count: 1
Volume1.arbiter_count: 1
Volume1.disperse_count: 0
Volume1.redundancy_count: 0
Volume1.quorum_status: not_applicable
Volume1.snapd_svc.online_status: Offline
Volume1.snapd_svc.inited: True
Volume1.rebalance.id: 00000000-0000-0000-0000-000000000000
Volume1.rebalance.status: not_started
Volume1.rebalance.failures: 0
Volume1.rebalance.skipped: 0
Volume1.rebalance.lookedup: 0
Volume1.rebalance.files: 0
Volume1.rebalance.data: 0Bytes
Volume1.options.transport.address-family: inet
Volume1.options.performance.readdir-ahead: on
Volume1.options.nfs.disable: on
```

@samikshan can we add that detail here?

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,275 @@
...
+
+* Identify a arbiter volume and mark accordingly in central store, based on
+field `volume{n}.arbiter_count` in `get-state` output
+
+* Remove the uses of `gstatus` and use aggregations across bricks of the volume
+to decide the volume utilization data
+
+* Remove the uses of `gstatus` and use aggregations of `gluster peer status` and
+`gluster volume status` to decide the cluster health status
+
```
r0h4n: Each volume{n},  should be synced on a new greenlet

r0h4n (changes requested):

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Ack .  Will take care while code changes....

specs/enhance_gluster_cluster_sync_logic.adoc
atinmu: arbiter brick will be always the last brick (3rd to be precise) in the chain.

shtripat: As discussed today it would be taken care in `get-state` output. The github issue created fro the same is at https://github.com/gluster/glusterfs/issues/278

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
+= Enhance gluster cluster synchronization logic as per recommendations
+
+The intent of this change is to accommodate additional details available as part
+of `get-state` output while syncing the gluster cluster details and store them
+in central store.
+
+Also instead of using `gstatus` for cluster health and volume utilization
```
cybernth, line 7: By implementing this spec, I assume that we are removing the dependency with gstatus? if that is the case we need to account for the spec file changes as well

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
...
+* Volume snapshot details
+* Add volume re-balance estimation time details
+* Volume geo-rep details
+* Client connections per volume
+* Identification of arbiter volumes in cluster (if any)
+* Instead of `gstatus` use aggregations across bricks utilization to calculate
+volume utilization
+* Instead of `gstatus` use peer status and volume status to aggregate and
+calculate the cluster status
+
```
cybernth, line 38: How we are planning to calculate volume status? I think `gluster volume status` gives the info whether the volume is started or not? gstatus gives states like `degraded` etc. We need to have something similar

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+== Proposed change
+
+* Use `get-state detail` and `get-state volumeoptions` commands instead of plain
```
cybernth, line 42: Can you add sample outputs for `get-state detail` and `get-state volumeoptions` commands

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+* Identify a arbiter volume and mark accordingly in central store, based on
+field `volume{n}.arbiter_count` in `get-state` output
+
+* Remove the uses of `gstatus` and use aggregations across bricks of the volume
+to decide the volume utilization data
+
+* Remove the uses of `gstatus` and use aggregations of `gluster peer status` and
+`gluster volume status` to decide the cluster health status
+
```
cybernth: What about :
1. Heal information
2. What are the different volume types we support now? do we support all?

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
...
+```
+
+* Add an additional field `rebal_estimation_time` under `RebalanceDetails` to
+hold the re-balance estimation time value
+
+* Add an additional object `GeoRepSessions` to hold the geo-rep session details
+for gluster volumes
+
+```
+TBD
```
cybernth: please add this info

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,307 @@
...
+the client connections details
+
+* Add ad additional fields `client_connections` and `arbiter` under `Brick`
+object to hold the client connections and to mark if the brick is arbiter brick
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* Provide an API to list snapshot details for volumes
```
cybernth: Why do you think these APIs are required?. We are not going to do day1,day2 now

cybernth (changes requested):

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Yes we remove the dependency of `gstatus` with these changes. Will add the spec change as well.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Volume status is something which is returned in `get-state` and we will use that as earlier. Volume state will be calculated using the same logic which `gstatus` had to form the cluster status value. Code changes PR would give more idea (actually the logic is copied from `gstatus`).

Do you need volume status as degraded etc? Earlier also we were using only `get-state` output for volume status. The volume state was used by `gstatus` only for deciding the cluster status (healthy/unhealthy)

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: `get-state detail` output is there at bottom of the spec. Will add a sample of `get-state volumeoptions` as well.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Everything which is available by `get-state detail` and `get-state volumeoptions` is covered as part of this spec. As part of this we have `estimated_time_left` for rebalance though which is calculated based on suggestion by gluster team (maximum of `time_left` reported from different nodes for rebalance for a volume).

Should we run `heal info` as well for each volume and sync details? I understand heal should be enabled/disabled also at volume level.

For cluster status and utilization all the volume types are supported actually and the logic is copied from `gstatus`

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Done. This would be added once we have geo-rep detail available in `get-state` which is planned.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Listing API should be available I feel.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+
+
+== Proposed change
+
+* Use `get-state detail` and `get-state volumeoptions` commands instead of plain
+`get-state` as first provides the detailed (e.g. client connections etc.)
+information for the volume and second is option to list all the volume options
+with default values
+
+* Use `get-state` output to figure out the volume snapshot details and update in
```
brainfunked, line 47: What are the details and how would they be calculated? The logic needs to be documented in the spec and verified before implementing in code.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+central store
+
+* Calculate the estimation time for re-balance job. This should be largest
+value among the `time_left` values reported by get-state from different nodes of
+the cluster
+
+* Use `get-state` output for geo-rep session details like no. of geo-rep
+sessions and active geo-rep sessions and update per volume in central store
```
brainfunked, line 57: Are the details readily available in the `get-state` output or does some logic need to be applied?

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+
+* Figure out the client connections per brick from `get-state` output and
+aggregate for volume and update in central store
```
brainfunked: Please document the aggregation strategy.

Also, I think there are certain pitfalls in using the client connections provided by gluster, such as no differentiation between actual clients and other processes. This needs to be documented clearly in the spec as it would be necessary for the documentation of this feature.

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: @shtripat please document these details in the spec itself.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+
+* Identify a arbiter volume and mark accordingly in central store, based on
+field `volume{n}.arbiter_count` in `get-state` output
+
+* Remove the uses of `gstatus` and use aggregations across bricks of the volume
+to decide the volume utilization data
```
brainfunked: Please document the aggregation logic.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+
+* Remove the uses of `gstatus` and use aggregations of `gluster peer status` and
+`gluster volume status` to decide the cluster health status
```
brainfunked: What's the aggregation logic?

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: Document in the spec please.

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: @atinmu what's your take on this?

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,375 @@
...
+Primary assignee:
+  shtripat
+  nnDarshan
+
+=== Work Items:
+
+* https://github.com/Tendrl/specifications/issues/167
+
+
+== Dependencies:
```
brainfunked, line 406: I believe there are some changes to the `get-state` that need to be made in gluster upstream. All the gluster upstream issues tracking these changes need to be listed as dependencies. @atinmu may be able to help with the list of these issues.

brainfunked (changes requested):

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: There is no logic involved in getting snapshot details from `get-state` output. All the snapshots list is available in `get-state` output and same would be captured and instances of class `Snapshot` (defined in data model section) would be created and persisted in central store.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Gluster team is not yet ready with geo-rep details in `get-state` output. Once its available details of the structure of object etc and if any aggregations would be documented in this spec.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Aggregation logic for client connections is simple. Simply add all the client connection for the participating bricks and that makes the value for volume.
Will add this details with notice that its as per guidelines from gluster team.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Sure

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Sure. would add details

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Dependencies section does list the items required from gluster team and also the references section has link to glusterfs github issues link. What else is missing here?

specs/enhance_gluster_cluster_sync_logic.adoc
cybernth: @anmolbabu , FYI Please make sure collectd plugins doesn't have dependency to gstatus going forward

anmolbabu: @nthomas-redhat Ack.. This has been done on existing plugins as in https://github.com/Tendrl/node-monitoring/pull/79 but yes that PR will eventually have to land on node-agent rather than node-monitoring(that its currently on)

specs/enhance_gluster_cluster_sync_logic.adoc
cybernth: Is there any specific reason why you maintain this separately? Can you list down the possible values for `volume state' and 'volume status' here? I think its better to have single state/status for volume

specs/enhance_gluster_cluster_sync_logic.adoc
cybernth: ack

specs/enhance_gluster_cluster_sync_logic.adoc
cybernth: Add this only if there is a specific requirement

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: The status and state are two different things. Volume might be degraded due to some replicas not available or it could be partially available. But still volume status is `Started`. I feel it fine to have like this.
Volume state is something which dont maintain in central store for a volume. We use this state data to figure out the cluster's healthy/unhealthy status only. Still you feel details required in spec, I can add all possible values.

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Ack. @anivargi ^^^

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: @nthomas-redhat As discussed based on bricks status and type of the volume, this logic reports 4 volume states which `up`, `down`, `up(partial)` and `up(degraded)`. The same values can be updated as additional field `state` under volume with this logic.

The volume `state` aggregated based on above logic is more detailed and can be used to display in UI.

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,492 @@
...
+** First calculate the `raw_capacity` and `raw_used` values for volume and
+cluster merely by adding the `raw_capacity` and `raw_used` for bricks of the
+volume. These details are synced and available with central store for the
+individual bricks
+
+** Now the pseudo code to decide the final values of volume and cluster usable
+and used capacity goes as below
+
+```
+if no of up bricks == no of total bricks:
```
shtripat, line 84: @atinmu below is the pseudo code for calculation of volume and cluster utilization. The logic is simple copy from `gstatus` at the moment. Need your inputs to figure if its correct or some changes would be required in this logic.
May be if required you can add others from Gluster team to review the logic as well.

specs/enhance_gluster_cluster_sync_logic.adoc
cybernth: ack

r0h4n (dismissed):

cybernth (dismissed):
Looks good

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: Does gdeploy also report the `up (partial)` status? I know it returns `up (degraded)`. I'm asking because if this is a new status that we report, it'll need to be documented to tell the users what it actually means.

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: Is there a gluster issue we could link to here?

specs/enhance_gluster_cluster_sync_logic.adoc
```diff
@@ -0,0 +1,494 @@
...
+* Use `get-state` output for geo-rep session details like no. of geo-rep
+sessions and active geo-rep sessions and update per volume in central store
+
+* Figure out the client connections per brick from `get-state` output and
+aggregate for volume and update in central store. The no of client connections
+for a volume merely a sum of all the participating brick's client connections.
+Note: This logic for calculating client connections for a volume is trivial but
+its as per the guidelines provided from Gluster team.
+
+* Identify a arbiter volume and mark accordingly in central store, based on
```
brainfunked: s/a/an

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: The APIs will soon be required, most likely in the [Gluster Monitoring milestone](https://github.com/Tendrl/specifications/milestone/1). Though, I suggest that for now, we exclude the API details from this spec and file a separate spec or an issue for API endpoints for all the additional gluster state information we now include as part of this spec.

specs/enhance_gluster_cluster_sync_logic.adoc
brainfunked: Would be ideal to link to the gluster issues tracking these changes.

brainfunked (commented):
Looks good for implementation, but I'd suggest addressing some minor concerns such as linking to gluster issues in this spec before merging it.

@r0h4n @nthomas-redhat please feel free to merge this spec, without awaiting an explicit ack from me, when the requested additions (if possible) are made to the document.

shtripat:
>> Looks good for implementation, but I'd suggest addressing some minor concerns such as linking to
>> gluster issues in this spec before merging it.

The references section does have links to glusterfs github issues.

shtripat:
>> Does gdeploy also report the up (partial) status? I know it returns up (degraded). I'm asking because if
>> this is a new status that we report, it'll need to be documented to tell the users what it actually means.

`up (partial)` and `up (degraded)` are two different states reported for volumes based on calculations. The pseudo code has the logic which decides in what situations the state would be `up (degraded)`

specs/enhance_gluster_cluster_sync_logic.adoc
shtripat: Done

shtripat:
>> The APIs will soon be required, most likely in the Gluster Monitoring milestone. Though, I suggest that
>> for now, we exclude the API details from this spec and file a separate spec or an issue for API endpoints
>> for all the additional gluster state information we now include as part of this spec.

Will remove the entry from this spec. @anivargi please do the needful from API side.

shtripat:
>> Would be ideal to link to the gluster issues tracking these changes.

I have gluster github issues in references section. Will move here.

cybernth (approved):

r0h4n (approved):

Issue #172, opened 2017-07-19
Performance enhancements for tendrl components


shtripat:
There are various aspects to performance enhancements for tendrl components like
- Clustered etcd setup for load balancing
- Instead of writing tendrl objects field by field to etcd, write as a single JSON and similarly while read read as a whole and then weave the object back
- Improvements on memory consumption
- Define standard configuration for the tendrl server node and additional requirement on storage nodes for tendrl components
- Response time tuning for REST endpoints
- If possible introduce some purging mechanism for log messages pushed into etcd (maintaining a separate etcd instance for audit log could be another option)

mbukatov:
Do I understand this right that when there is an spec pull request linked by github, I can ignore all those commits linked there? In other words, is it ok for me to review just https://github.com/Tendrl/specifications/pull/174 and ignore all the rest of commits and closed pull requests linked there?

PR #174, opened 2017-07-19
WIP - Added spec tendrl_performance_enhacements.adoc

shtripat:
tendrl-bug-id: Tendrl/specifications#172

### DIFF
--- a/specs/tendrl_performance_enhacements.adoc
+++ b/specs/tendrl_performance_enhacements.adoc
@@ -1,0 +1,235 @@
+= Tendrl performance enhancements for lesser CPU and memory consumption
+
+The intent of this change is to make sure load due to tendrl components on
+storage nodes is minimal. It also covers the aspects related to performant REST
+apis and make sure no crashes in etcd, predictable job processing with defined
+CPU and memory uses.
+
+It also tends to define the hardware requirements for standard tendrl server
+and load incurred on the storage nodes due to tendrl components.
+
+
+== Problem description
+
+This specification talk about various changes required in tendrl components to
+make it more performant and make sure they consume less resources (CPU, memory)
+on storage nodes. It also covers the guidelines to storage admin for required
+hardware for tendrl server, etcd clustering and load incurred on storage nodes.
+
+
+== Use Cases
+
+* This addresses the changes in the way tendrl entities get written and read
+to/from etcd. Currently the objects get written field by field which is CPU
+intensive and needs more resources.
+
+* The job processor in tendrl, consistently looks at `/queue` etcd dir for
+finding the jobs to be processed. We need a tagged job queue mechanism which
+reduces huge fetching and probing of the `/queue` jobs. With tagged job queues,
+specific services would look for the interesting specific job queues and they
+would process jobs them only.
+
+* Provide guidelines on standard hardware requirements for tendrl server node
+
+* Provide guidelines on setting up a clustered etcd for tendrl
+
+* Tuning of REST endpoints for better performance and predictable response time
+
+* Tuning of different components of tendrl for better memory utilizations
+
+
+== Proposed change
+
+* Annotate flows in tendrl definition files with tagged queue names (to which
+these flows would write the job to)
+
+* Introduce a tagged job queue mechanism in `tendrl-commons` module. Services
+with defined tags would pick jobs from their specific tagged job queues for
+processing
+
+* Enhance REST layer to create job in tagged job queues based on flow annotation
+for job queue names
+
+* Enhance writing/reading to/from etcd to consider whole object details as
+single JSON. While writing we need to get the json representation of the object
+and write as single field under etcd. While reading, it should be read as single
+value and whole object should be weaved back from JSON.
+
+A pseudo save and load functions would something like below
+
+```
+def save(self, update=True, ttl=None):
+    NS._int.wclient.write(self.value + '/data', self.json)
+
+def load(self):
+    self.render()
+    val = json.loads(NS._int.client.read(self.value + '/data').value)
+    for attr_name, attr_val in vars(self).iteritems():
+        if not attr_name starts with '_' and attr_name is not 'value':
+            Get attr type from definitions file already loaded
+            if attr type in ['json', 'list']:
+                setattr(self, attr_name, json.loads(attr_val))
+            else:
+                setattr(self, attr_name, attr_val)
+    return self
+```
+
+* Fine tune REST endpoints for better and faster response times
+
+* Document the hardware requirements for tendrl server under wiki
+
+* Document the clustering mechanism of etcd in wiki
+
+* Document the details of load incurred on storage nodes due to tendrl
+components within justified limits (so that storage admin can plan the resource
+requirements accordingly)
+
+* In {gluster/ceph}_integration, change the sds sync as a job and start this job
+while startup of these integration services. Once started, these jobs should be
+triggered periodically.
+
+* Change any explicit raw reads in tendrl components to use load() and then the
+required field from object.
+
+=== Alternatives
+
+* Regarding {gluster/ceph}_integration sds_sync as flows, there is another
+suggestion to have different time intervals at which specific details get
+synchr0nized. A sample pseudo code could be as below
+
+```
+counter = 1
+while True:
+    sleep(10)
+    sync volume and bricks
+
+    if counter % 30 == 0: # if 30 rounds of volume sync has happened, trigger
+        sync cluster status
+        sync utilization details
+
+    if counter % 60 == 0: # if 60 rounds of volume sync has happened, trigger
+        sync snapshots
+        sync underlying device details for bricks
+
+    counter = (counter + 1) % 60 (LCM of 1, 30, 60)
+```
+
+This would make sure different syncs are done at different intervals. Also this
+does not require sds_sync to be a separate flow and still different syncs get
+triggered at different intervals.
+
+=== Data model impact
+
+* Annotate the tendrl flows in different definitions files of tendrl modules to
+define the tagged queue name where these jobs would be written
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* With proposed changes above, the object details would be saved as single JSON
+field with name `data`. For example, the volume details would be saved as
+`clusters/{int-id}/Volumes/{vol-id}/data`. API layer need to change to read the
+values as per these changes for listing the objects details.
+
+* REST layer to write the jobs in tagged queues based on definitions
+
+* Enhancements for tuning the response time for various GET endpoints
+
+==== Notifications/Monitoring impact:
+None
+
+==== Tendrl/common impact:
+
+* Enhancements for processing tagged job queues. Based on the current service,
+it should look at defined tagged job queue only for figuring out the jobs to be
+picked and processed
+
+* Enhance the writing/reading logic to/from etcd to consider the whole object as
+single JSON
+
+==== Tendrl/node_agent impact:
+
+* Definitions changes for tagging flows with specific job queue names
+
+==== Sds integration impact:
+
+* Definitions changes for tagging flows with specific job queue names
+
+==== Tendrl Dashboard impact:
+
+None
+
+=== Security impact:
+
+None.
+
+=== Other end user impact:
+
+None
+
+=== Performance impact:
+
+None.
+
+=== Other deployer impact:
+
+None.
+
+=== Developer impact:
+
+None.
+
+
+== Implementation:
+
+* https://github.com/Tendrl/documentation/issues/88
+
+* https://github.com/Tendrl/documentation/issues/89
+
+* https://github.com/Tendrl/documentation/issues/90
+
+* https://github.com/Tendrl/commons/issues/657
+
+=== Assignee(s):
+
+Primary assignee:
+  shtripat
+  r0h4n
+  anivargi
+
+=== Work Items:
+
+* https://github.com/Tendrl/specifications/issues/172
+
+
+== Dependencies:
+
+None
+
+
+== Testing:
+
+* Verify that load incurred on storage nodes due to tendrl components is within
+the defined limits
+
+* Verify the REST endpoints for their response time and it should be within the
+defined time limits
+
+* Verify the guidelines published regarding clustering of etcd
+
+* Verify all the objects listing REST endpoints to make sure all the details are
+listed properly.
+
+
+== Documentation impact:
+
+* Document for clustered setup of etcd
+
+* Document for hardware requirements for tendrl server
+
+* Document for load details on storage nodes due to tendrl components
+
+== References:
+
+None


specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+=== Alternatives
+
+None
+
+=== Data model impact
+
+* Annotate the tendrl flows in different definitions files of tendrl modules to
+define the tagged queue name where these jobs would be written
+
```
r0h4n, line 125: No impact on tendrl code due to changes to BaseObject.save and load ?

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,172 @@
...
+* Fine tune REST endpoints for better and faster response times
+
+* Document the hardware requirements for tendrl server under wiki
+
+* Document the clustering mechanism of etcd in wiki
+
+* Document the details of load incurred on storage nodes due to tendrl
+components within justified limits (so that storage admin can plan the resource
+requirements accordingly)
+
```
r0h4n, line 86: Add item:
- For gluster and ceph sds sync, convert the sync thread to a job and start the job at integration service startup and then triggered periodically. 
- after the above change, each sds sync object (eg: volume) should be synced in a separate greenlet.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,172 @@
...
+* Annotate the tendrl flows in different definitions files of tendrl modules to
+define the tagged queue name where these jobs would be written
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+* REST layer to write the jobs in tagged queues based on definitions
+
+* Enhancements for tuning the response time for various GET endpoints
```
r0h4n: The changes to BaseObject ("data" attr) will have an API impact, please document it

r0h4n (changes requested):

specs/tendrl_performance_enhacements.adoc
shtripat: Ok. Will add the required details

specs/tendrl_performance_enhacements.adoc
shtripat: Haven't tried yet. I feel job creations which are done explicitly from code might get impacted.
I think you had some design in mind. You can add details as well.

specs/tendrl_performance_enhacements.adoc
shtripat: Yes. Will add a dependency on API layer.

shtripat:
@anivargi kindly review the API dependencies and section specific to it.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,184 @@
...
+finding the jobs to be processed. We need a tagged job queue mechanism which
+reduces huge fetching and probing of the `/queue` jobs. With tagged job queues,
+specific services would look for the interesting specific job queues and they
+would process jobs them only.
+
+* Provide guidelines on standard hardware requirements for tendrl server node
+
+* Provide guidelines on setting up a clustered etcd for tendrl
+
+* Tuning of REST endpoints for better performance and predictable response time
```
anivargi, line 36: What does this exactly involve?

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,184 @@
...
+== Proposed change
+
+* Annotate flows in tendrl definition files with tagged queue names (to which
+these flows would write the job to)
+
+* Introduce a tagged job queue mechanism in `tendrl-commons` module. Services
+with defined tags would pick jobs from their specific tagged job queues for
+processing
+
+* Enhance REST layer to create job in tagged job queues based on flow annotation
```
anivargi, line 50: We are already adding tags for jobs.

anivargi (changes requested):

specs/tendrl_performance_enhacements.adoc
shtripat: I remember while initial discussions we talked about tuning `GetNodeList` kind of APIs which were taking lot of time to load details. You can comment more if such things to be done.

specs/tendrl_performance_enhacements.adoc
shtripat: We are adding tags. We would need changes to submit the jobs to tagged queues instead of putting to `/queue` always. That's the change being talked about here.

shtripat:
@r0h4n I have pushed the code changes PRs for commons/node-agent for this.
You would need to push your changes which you have done for performance enhancements under this spec (specifically for tagged job queues and others)

shtripat:
@nnDarshan I have the added the alternate mechanism which you suggested for making different sync calls at different intervals. Plz have a look.

@r0h4n plz check and comment.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+* Provide guidelines on setting up a clustered etcd for tendrl
+
+* Tuning of REST endpoints for better performance and predictable response time
+
+* Tuning of different components of tendrl for better memory utilizations
+
+
+== Proposed change
+
+* Annotate flows in tendrl definition files with tagged queue names (to which
```
cybernth, line 43: Can you provide the details here? What are the different tagged queues and the flows attached to it? Also provide sample definition file with this change.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+these flows would write the job to)
+
+* Introduce a tagged job queue mechanism in `tendrl-commons` module. Services
```
cybernth, line 46: Provide specifics/psuedocode about the proposed change

specs/tendrl_performance_enhacements.adoc
cybernth: @anivargi , @shtripat , we need this details in the docs. what are the chnages required on API side. Import flows needs to taken care of immediately and what else? I guess queue(to submit the job) related information to be read from the definition files

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+with defined tags would pick jobs from their specific tagged job queues for
+processing
+
+* Enhance REST layer to create job in tagged job queues based on flow annotation
+for job queue names
+
+* Enhance writing/reading to/from etcd to consider whole object details as
+single JSON. While writing we need to get the json representation of the object
+and write as single field under etcd. While reading, it should be read as single
+value and whole object should be weaved back from JSON.
```
cybernth, line 56: Provide the psuedocode for this change

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+* Fine tune REST endpoints for better and faster response times
```
cybernth: do we have a list of APIs undergo change? what changes are required. I think this should be covered in a different spec

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+* Document the hardware requirements for tendrl server under wiki
+
+* Document the clustering mechanism of etcd in wiki
+
+* Document the details of load incurred on storage nodes due to tendrl
+components within justified limits (so that storage admin can plan the resource
+requirements accordingly)
```
cybernth, line 85: Lines 60-66 - where this documention available? Are we spec - ing this out here or in a different spec?

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+* In {gluster/ceph}_integration, change the sds sync as a job and start this job
+while startup of these integration services. Once started, these jobs should be
+triggered periodically.
```
cybernth, line 89: Psuedocode required

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+* Change any explicit raw reads in tendrl components to use load() and then the
+required field from object.
```
cybernth, line 92: Psuedocode required

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,210 @@
...
+specific services would look for the interesting specific job queues and they
+would process jobs them only.
+
+* Provide guidelines on standard hardware requirements for tendrl server node
+
+* Provide guidelines on setting up a clustered etcd for tendrl
+
+* Tuning of REST endpoints for better performance and predictable response time
+
+* Tuning of different components of tendrl for better memory utilizations
```
cybernth, line 38: What are changes proposed to address this?

cybernth (changes requested):

specs/tendrl_performance_enhacements.adoc
anivargi: @shtripat it all depends on the etcd structure to what extent the endpoints be optimized, if etcd can maintain a index of how the objects are updated then we can do some intelligent caching on the API level. Also having dynamic (Memory) and static(os version) attributes separated would also help in improving the response time. 
We already got a improvement in nodes list by removing the monitoring data from the response.

specs/tendrl_performance_enhacements.adoc
anivargi: I think the queue changes needs to be a different spec.

specs/tendrl_performance_enhacements.adoc
shtripat: @anivargi you are the best person to comment here.

specs/tendrl_performance_enhacements.adoc
shtripat: @r0h4n may be you can start another spec linked to this as you have already thought through a lot around this.

specs/tendrl_performance_enhacements.adoc
shtripat: @r0h4n ^^^

specs/tendrl_performance_enhacements.adoc
shtripat: Isn't this logic very trivial in nature? Anyway will add sample.

specs/tendrl_performance_enhacements.adoc
shtripat: @r0h4n thoughts?

specs/tendrl_performance_enhacements.adoc
shtripat: As part this spec only I have listed so we don't forget this. All we need is some wiki docs written for these.
If you want can create individual issues against the `Documentation` repo

specs/tendrl_performance_enhacements.adoc
shtripat: This was suggested recently. Also portions of this spec are to be implemented/already available from @r0h4n. I would request @r0h4n also to add flesh to this spec.

specs/tendrl_performance_enhacements.adoc
shtripat: This is for sure very trivial and I strongly feel there is no need for pseudo code.
All we would be doing is instead os `NS._int.client.read` we would be using `NS.{mod ns}.objects.{Class}.load()`.

Do you feel we need a pseudo code for these kind of trivial things in spec?

specs/tendrl_performance_enhacements.adoc
brainfunked: Not sure what this has to do with only the API. Wouldn't this need to be done for all the backend components as well?

specs/tendrl_performance_enhacements.adoc
brainfunked: I agree, should be a different spec.

brainfunked (changes requested):
Best to separate out the changes that impact the API into a separate spec issue and remove them from here. Correspondingly, I think this spec issue and the spec itself could be renamed to something more specific.

The new spec could be part of the [Gluster Monitoring milestone](https://github.com/Tendrl/specifications/milestone/1) @anivargi @nthomas-redhat @r0h4n

specs/tendrl_performance_enhacements.adoc
shtripat: Ok. based on this discussion I would remove this line which is very specific to API layer. All backend changes anyway are listed in the spec here.

specs/tendrl_performance_enhacements.adoc
shtripat: @r0h4n ^^^

shtripat:
>> Not sure what this has to do with only the API. Wouldn't this need to be done for all the backend
>> components as well?

This is point under use cases and not specific to API only.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+
+== Dependencies:
+
+None
+
+
+== Testing:
+
+* Verify that load incurred on storage nodes due to tendrl components is within
```
mkudlej, line 213: What does mean defined limits in this context? I mean what is goal here? I think this should be defined here and if it cannot be achieved with specified HW configuration then result should be new limits with reasons why limits have been changed.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,235 @@
...
+the defined limits
+
+* Verify the REST endpoints for their response time and it should be within the
```
mkudlej, line 216: Similar question than above.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+* The job processor in tendrl, consistently looks at `/queue` etcd dir for
+finding the jobs to be processed. We need a tagged job queue mechanism which
+reduces huge fetching and probing of the `/queue` jobs. With tagged job queues,
+specific services would look for the interesting specific job queues and they
+would process jobs them only.
+
+* Provide guidelines on standard hardware requirements for tendrl server node
+
+* Provide guidelines on setting up a clustered etcd for tendrl
```
mkudlej, line 34: Are Tendrl components ready for clustered Etcd? Because right now there is only one IP/FQDN for accessing Etcd in configuration.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+* Verify the guidelines published regarding clustering of etcd
+
+* Verify all the objects listing REST endpoints to make sure all the details are
+listed properly.
+
+
+== Documentation impact:
+
+* Document for clustered setup of etcd
```
mbukatov, line 227: Do we really consider clustered etcd setup for the [milestone #3](https://github.com/Tendrl/specifications/milestone/3)? I'm asking because such change will have a huge impact on setup and configuration in general.

specs/tendrl_performance_enhacements.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+* Document for hardware requirements for tendrl server
+
+* Document for load details on storage nodes due to tendrl components
```
mbukatov, line 231: What does this mean exactly? Does it refer to expected performance impact of tendrl components on storage nodes?

specs/tendrl_performance_enhacements.adoc
mbukatov: After discussion with dev team, this was moved out into [3rd milestone](https://github.com/Tendrl/specifications/milestone/1).

Issue #190, opened 2017-07-19
Alert notifications


rishubhjain:
Rename the current alerting module(https://github.com/Tendrl/alerting) to Notifier.
Check if any changes are required to trim the module down to:
- Monitoring the alert conditions in etcd.
- Sending alert notifications to configurable receiving end-points: ui and smtp.

mbukatov:
Which is the pull request with latest specification document I should review to see documentation (interested in configuration and setup) to understand changes needed in tendrl-ansible? I wasn't able to find the details I'm interested in.

I'm going to assume that all details I need to be aware of are described in: https://github.com/Tendrl/tendrl-ansible/issues/33

PR #198, opened 2017-07-20
Grafana event handling in monitoring-integration spec

rishubhjain:
tendrl-bug-id: Tendrl/specifications#169

PR #199, opened 2017-07-20
WIP - Spec for Alert_Notifications

rishubhjain:
tendrl-bug-id: Tendrl/specifications#190

### DIFF
--- a/specs/Alert_Notifications.adoc
+++ b/specs/Alert_Notifications.adoc
@@ -1,0 +1,200 @@
+
+
+= Alert Notifications
+
+Alerting system is essential to spot the problem and notify it to the corresponding user.
+The thresholds for alerts are configured in grafana and events of alerts will be notified by grafana.
+These events are interpretted and modified by Monitoring-Integration component and notified to
+respective user via the notifier.
+
+
+== Problem description
+
+The present alerting system monitors all the events in etcd and retrieves alert specific events.
+The events are converted into alert structure and stored in etcd under /alerting/alerts/.
+This leads to duplication of events in etcd and excess read and writes to etcd.
+The events are processed for alerts after saving into etcd instead of processing it before
+saving events in etcd.
+
+== Use Cases
+
+Goal: Notify user in case of any alert event.
+
+Actors:
+
+1) Notifier
+
+Steps:
+
+* Node-Agent stores events in etcd.
+
+* The alert conditions will be monitored by the Notifier.
+
+* The alert notification is sent to the appropriate end points via
+  corresponding notification channels.
+
+== Proposed change
+
+Present Alert system(https://github.com/Tendrl/alerting) will be renamed as “Notifier”
+
+The main functionality of Notifier would be to:
+
+* Monitor etcd for alert conditions being set, modified or unset.
+
+* Sending alert notifications to configurable receiving end-points.
+
+Presently the alerting module processes all events under messages\events
+in etcd and looks for events having priority as “notice” which signifies they are
+alerts. Instead of processing all events an AlertManager will be configured to
+receive just alerting events and notify users about these events.
+
+For this the notifier must receive only the alerting events.
+This can be achieved in 2 ways:
+
+* All alert conditions will be stored in etcd under alerting/alerts/{cluster,node} by
+  the node-agent as per the scope of the alert condition (cluster level scope or node level scope).
+  A watcher(NotificationWatcher) will be configured to poll etcd (alerting/alerts) for any new
+  set or unset alert conditions. On detecting new alert condition or changes in existing condition
+  the alert condition will be passed to notifier and using the configured notification channel the
+  notifier would notify the users of these alerts.
+
+ Advantage:
+
+     ** It will reduce number of read and write in etcd.
+
+  Drawback:
+
+      ** Polling etcd for new alerts is not efficient.
+
+* To avoid polling etcd, node-agent instead of just logging the events to
+  alerting/alerts/{cluster,node} could also pass events to notifier directly.
+  Notifier can extract the payload,parse it and using the configured notification channel
+  would notify the users of these events.
+
+    Advantage:
+
+     ** Avoids polling etcd for new alerts.
+
+
+The Notifier will notify the user of the events via configured notification channels
+like SMTP.
+
+=== Alternatives:
+
+Prometheus alertmanager can be used as a "Notifier" instead of tendrl alert module.
+It provides facility to inhibit, group and silence the alerts.
+
+=== Data model impact:
+
+ No changes in existing structure.
+
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+None
+
+==== Notifications/Monitoring impact:
+
+None
+
+==== Tendrl/common impact:
+
+None
+
+==== Tendrl/node_agent impact:
+
+WIP
+
+==== Sds integration impact:
+
+None
+
+==== Tendrl/performance-monitoring impact:
+
+None
+
+==== Tendrl/alerting impact:
+
+Tendrl alerting system will be renamed as Notifier.
+Its main functionality would be to fetch the alert events and
+notify the respective user.
+
+=== Security impact:
+
+ None
+
+=== Other end user impact:
+
+Users will be able to set alerts in Grafana itself and get notified for the same.
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+None
+
+=== Developer impact:
+
+Modify node-agent to insert events into different path in etcd.
+
+Trim down present alerting system to provide only notification functionality.
+
+== Implementation:
+
+* Rename the alerting module as notifier
+* Rename the file tendrl-alerting.spec as tendrl-notifier.spec
+* Rename the file tendrl-alerting.service as tendrl-notifier.service
+* Replace the word tendrl-alerting or alerting or both from tendrl-alerting.spec,
+  tendrl-alerting.service, setup.py, README.rst, Makefile, MANIFEST.in as tendrl-alerting
+  or alerting.
+* Rename the alerting_logging.yaml.sample as notifier_logging.yaml.sample and modifiy the
+  file.
+* Rename the file alerting.conf.yaml.sample as notifier.conf.yaml.sample and modify the
+  file.
+* Modify the doc/sources based on new notifier module functionality.
+* Modify the __init__ function in TendrlAlertingManager class to start gevent for watcher,
+  and sync thread only.
+* Rename the class TendrlAlertingManager as TendrlNotificationManager in __init__.py
+* Modify the AlertingNS as NotifierNS
+* Remove the handlers module from notifer.
+* Change the watcher to monitor alerts from /alerting/alerts instead of /messages/event
+* Create the file called alert_condition_handler.py in node-agent.
+* Create a new class called AlertCondtionHandler in alert_condition_handler.py
+* Move the alert object, cluster_alert object and node_alert from alerting module to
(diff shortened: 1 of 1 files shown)


specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,186 @@
+
```
shtripat, line 1: This spec could be referenced from https://github.com/Tendrl/specifications/pull/198 ?

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,186 @@
...
+   collect the events from etcd and then using the configs in etcd it could 
+   notify the respective user.
+
+ Advantage:
+ 
+     ** Simple architecture and easy to implement.
+
+  Drawback:
+ 
+      ** Pooling etcd for new alerts is not efficient.
```
shtripat: s/Pooling/polling/

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,186 @@
...
+ 
+* The node-agent can directly pass the alert events to the notifier via an endpoint
+   exposed by the notifier.
+
+    Advantage:
+ 
+     ** Avoids pooling etcd for new alerts.
```
shtripat: Same as above

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+== Implementation:
+
+WIP
+
+=== Assignee(s):
+
+
+Gowtham S
+Rishubh jain
```
shtripat: I dont see a mention about clearing alerts (say like we raised and alert saying CPU uses at the moment is 90%. But later when it goes down there should be a clearing alert for this existing one. This alert should vanish/or should not keep listing in tendrl UI).

If clearing is not provided user would manually need to correlate between threshold cross and then within limits cases.

specs/Alert_Notifications.adoc
rishubhjain: Done

specs/Alert_Notifications.adoc
rishubhjain: Done

r0h4n (approved):

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+Goal: Notify user in case of any alert event.
+
+Actors:
+
+1) User
+
+Steps:
+
+* Node Agent stores alerting events in etcd.
```
brainfunked: Should be: `Node agent sets the alert condition in etcd.`

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+* These events will be fetched by the Notifier.
```
brainfunked: Should be: `The alert conditions would be monitored by the Notifier.`

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+* The alerting event is to be sent to respective user via configured
+  notification channels.
```
brainfunked: Should be: `The alert notification is sent to the appropriate end points via corresponding notification channels.`

Alerts may not always end up directly with a user, they may be delivered to an intermediate system, eg. snmpd.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+== Proposed change
+
+
+Present Alert system(https://github.com/Tendrl/alerting) will be renamed as “Notifier”
+
+The main functionality of Notifier would be to:
+
+Monitor etcd only for alert specific events.
```
brainfunked: Should be: `Monitor etcd for alert conditions being set, modified or unset.`

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+* The alerting event is to be sent to respective user via configured
+  notification channels.
+
+== Proposed change
+
+
+Present Alert system(https://github.com/Tendrl/alerting) will be renamed as “Notifier”
+
+The main functionality of Notifier would be to:
```
brainfunked: Please use bullet list for all the items under this heading.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+Presently the alerting module processes all events under messages\events
+in etcd and looks for events having priority as “notice” which signifies they are
+alerts. Instead of processing all events an AlertManager will be configured to
+receive just alerting events and notify users about these events.
+
+For this the notifier must receive only the alerting events.
+This can be achieved in 2 ways:
+
+* All alerting events will be stored in etcd under alerting/alerts/{cluster,node} by
```
brainfunked: alert conditions, not events.

Also, not sure if we need the directory structure to be `alerting/alerts/<something>`. Is there anything else stored in the alerting directory apart from the alerts?

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+  the node-agent as per the scope of the event (cluster level scope or node level scope).
```
brainfunked: Again, alert condition, not event. Events never reach the node agent, they're processed at source and turned into an alert condition before handing to the node agent.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+  A watcher(AlertWatcher) will be configured to poll etcd (alerting/alerts) for any new
+  events. On detecting new events the evetns will be passed to notifier which would
```
brainfunked: Not just new conditions. It has to watch for changes to existing conditions, including unsetting. Any change to an alert condition state could be a possible notification.

Also, typo for `events` on this line.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+ Advantage:
+ 
+     ** Simple architecture and easy to implement.
+
+  Drawback:
+ 
+      ** Polling etcd for new alerts is not efficient.
+ 
+* To avoid polling etcd, node agent instead of just logging the events to
```
brainfunked: The logic behind setting the alert conditions first is that it is vital to ensure that the system knows about a condition before notifying it to the end user. Routing this via the notifier creates an extra hop in this loop and introduces a higher chance of the condition being lost.

In addition, the notifier's job is to only notify. It must not have write access to etcd in an important global directory. That's why it is the node agent's responsibility, as the transport for all the alert conditions, to set the alert condition first and foremost.

For this particular logic to work, the node agent would have to both set the alert condition and hand the said condition to the notifier.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+ 
+     ** Avoids polling etcd for new alerts.
+
+
+The Notifier will notify the user of the events via configured notification channels
+like SMTP.
+
+=== Alternatives:
+
+None
```
brainfunked: Prometheus alertmanager qualifies as an alternative, I think.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+==== Tendrl/common impact:
+
+None
+
+==== Tendrl/node_agent impact:
+
+Node-agent must be modified to insert the events received by Monitoring
+Integration in etcd under alerting/alerts/ instead of messages/events/.
+
+Monitoring integration adds scope ( global, cluster specific or node specific)
```
brainfunked: Monitoring integration is just one of the components that generates an alert condition from an event. Such would also be the case for node agent, {ceph,gluster}-integration. All of these systems are responsible for receiving events from their underlying systems.

brainfunked (changes requested):
The spec does not contain any information regarding detecting changes to existing alert conditions and generating notifications for the same. Notifying of new alert conditions, while essential, isn't the only goal of this system.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+Present implementation causes the alerting component to process events even
+when it can be done by a separate component before inserting the event in etcd.
+
+== Use Cases
+
+Goal: Notify user in case of any alert event.
+
+Actors:
+
+1) User
```
brainfunked: In the context of this spec, the actor is the Notifier and not the user. The user is just a passive receiver, who is never directly involved.

specs/Alert_Notifications.adoc
```diff
@@ -0,0 +1,189 @@
...
+Steps:
+
+* Node Agent stores alerting events in etcd.
+
+* These events will be fetched by the Notifier.
+
+* The alerting event is to be sent to respective user via configured
+  notification channels.
+
+== Proposed change
```
brainfunked: Deployment guidelines are missing.

specs/Alert_Notifications.adoc
GowthamShanmugam: i will add this in reference section

specs/Alert_Notifications.adoc
GowthamShanmugam: i will change this user to notifier

specs/Alert_Notifications.adoc
GowthamShanmugam: i will change this

specs/Alert_Notifications.adoc
GowthamShanmugam: i will change this

specs/Alert_Notifications.adoc
GowthamShanmugam: i will change this

specs/Alert_Notifications.adoc
GowthamShanmugam: i will add some points

specs/Alert_Notifications.adoc
GowthamShanmugam: i will add it

specs/Alert_Notifications.adoc
GowthamShanmugam: done

specs/Alert_Notifications.adoc
GowthamShanmugam: no,  as per flows which we proposed in https://github.com/Tendrl/specifications/pull/198 we are going to store alerts only

specs/Alert_Notifications.adoc
GowthamShanmugam: but node-agent will receive message-structure only via socket righ?


Here is the confusion is when monitoring-integration catch the grafana event it has to form tendrl alert structure using grafana alert and push that alert condition into message payload and passed into message socket right?

what @shtripat suggested is instead of node-agent push the alert condition into /alerting/alerts, node-agent only do log the alert messages into /messages/event, monitoring integration should pass the alert condition into /alerting/alerts. because when alert came it is stored into /alerting/alerts when clearing alert came we have to search the previous alert to update with this alert so doing this process in node-agent logger is not correct. 

so monitoring integration will push the message in node-agent socket as well as /alerting/alerts.

is this ok for this monitoring-integration can't be localhost. 
one more problem in string via node-agent is again logger has to convert alert structure in payload into alert object then only it can store. so alert object should be in common for this.

GowthamShanmugam:
@brainfunked @shtripat @anmolbabu changes done

GowthamShanmugam:
@brainfunked @anmolbabu @shtripat please review

GowthamShanmugam:
@r0h4n implementation is done for this spec we can close this

Issue #173, opened 2017-07-19
Sync underlying details of gluster brick.


nnDarshan:
As of now, gluster integration syncs only the brick details provided by the `gluster get state` command which is limited to mount point of the brick. Tendrl should gather the underlying device details of the brick which is of more interest to the storage admin.
This detail is something that does not change very often, so tendrl should have an efficient sync strategy for syncing these details. So that it does not hamper the performance.

nnDarshan:
The patch is merged

PR #201, opened 2017-07-21
Added spec for collecting gluster brick device detail

nnDarshan:
tendrl-bug-id: Tendrl/specifications#173
Signed-off-by: nnDarshan <darshan.n.2024@gmail.com>

### DIFF
--- a/specs/collect_brick_device_details.adoc
+++ b/specs/collect_brick_device_details.adoc
@@ -1,0 +1,120 @@
+= Collect Brick device details
+
+The underlying device details of gluster bricks has to be maintained in tendrl, as it helps
+admins to take corrective actions on the devices when the brick goes bad.
+
+
+== Problem description
+
+Gluster is just aware of the mount point that is used to create the bricks, it does not worry
+about the underlying devices as it operates with mountpoints. But if something goes wrong with
+the brick, its not of much help if tendrl reports that there is problem with the mount point
+as it is a logical entity and the root cause for this would be an underlying device. So tendrl
+should maintain the mapping between the bricks and underlying block devices.
+
+Each brick will map to one or more block devices, the name of block devices that are under
+the brick are stored as attribute of the brick `clusters/<cluster-id>/Bricks/all/<brick-path>/devices`.
+Each of the device in the list can be mapped to block devices under every node details, the
+path would be `nodes/<node-id>/LocalStorage/BlockDevices/all/<device_path>`
+
+== Use Cases
+
+* It helps the storage admin to easily take corrective actions when some thing happens to the
+Brick.
+
+== Proposed change
+
+Since brick device details are not provided by gluster get state. Tendrl should explicitly
+gather these details by using the mount point of the brick.
+
+For gathering this information tendrl can use blivet python module. Blivet is a python module
+for system storage configuration. Using blivet tendrl can gather the device hierarcy that is
+used to create the brick. Sample snippet of how to gather the information from blivet is given
+below:
+[source,python]
+----
+import blivet
+
+b = blivet.Blivet()
+b.reset() # this will gather the device information in this machine
+device = b.devicetree.resolveDevice(<filesystem>)
+
+ancestors = device.ancestors
+----
+
+=== Alternatives
+
+None
+
+=== Data model impact
+
+Data model remains the same, these details will be stored in the brick object, whose
+path is '/clusters/<cluster-id>/bricks/'
+
+=== REST API impact
+
+None
+
+=== Security impact
+
+None
+
+=== Notifications/Monitoring impact
+
+While there is a brick state change event, when we send the alerts we should also
+include the brick device information in the message.
+
+=== Other end user impact
+
+User will be able to view the underlying physical device info of a brick.
+
+=== Performance Impact
+
+None
+
+=== Other deployer impact
+
+None
+
+=== Developer impact
+
+None
+
+== Implementation
+
+=== Assignee(s)
+
+nnDarshan
+
+Primary assignee:
+nnDarshan
+
+Other contributors:
+None
+
+=== Work Items
+
+* gluster state sync should have additional piece of code to get the brick device details
+using blivet.
+
+* the updation of device details in the sds state sync thread should happen via a seperate
+greenlet
+
+== Dependencies
+
+This spec would add a new rpm(python-blivet) depencency on gluster-integraion
+
+== Testing
+
+Make sure that the underlying device details are properly synced for different types of
+devices.
+
+== Documentation Impact
+
+None
+
+== References
+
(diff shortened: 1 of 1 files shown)


nnDarshan:
@Tendrl/tendrl-core @shtripat Please review

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+Primary assignee:
+nnDarshan
+
+Other contributors:
+None
+
+=== Work Items
+
+* gluster state sync should have additional piece of code to get the brick device details
```
r0h4n, line 97: I would want to split such per volume state sync to another greenlet spawned from inside state sync thread.

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,113 @@
...
+using blivet.
+
+== Dependencies
```
r0h4n: python-blivet available for epel/centos7/rhel ?

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+=== Security impact
+
+None
+
+=== Notifications/Monitoring impact
+
+While there is a brick state change event, when we send the alerts we should also
+include the brick device information in the message.
+
```
r0h4n, line 66: There should be an alert against the brick and volume if this device behind the mountpoint goes down or fails

r0h4n (changes requested):

specs/collect_brick_device_details.adoc
nnDarshan: Detecting device failures is out of scope of this spec. This spec will just find the device hierarchy of the glusterfs bricks.
We will have to explore more tools to find a way to detect device failures. Which I feel is a candidate for another spec

specs/collect_brick_device_details.adoc
nnDarshan: ack, will add that requirement in spec

specs/collect_brick_device_details.adoc
nnDarshan: yes, its available in cenos base and rhel

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,116 @@
...
+ancestors = device.ancestors
+----
+
+=== Alternatives
+
+None
+
+=== Data model impact
+
+Data model remains the same, these details will be stored in the brick object, whose
```
shtripat, line 51: Don't we need  additional field under `Brick` object fir these device details? If so please add details of those fields.

specs/collect_brick_device_details.adoc
shtripat: Each volume sync is a greenlet as part of https://github.com/Tendrl/gluster-integration/pull/347 anyway.

specs/collect_brick_device_details.adoc
nnDarshan: These were already there but. It was updated only when we provisioned bricks through tendrl. With this spec we are trying to update these details as part of sync. So, no change to brick data model

shtripat: Oh ok. I didnt realize that. It fine.

brainfunked (commented):
Need some clarification on this spec.

I'd imagine that monitoring of system storage (mount points, block devices etc.) would fall under the node agent's disk inventory sync. In the scenario where a mounted block device is used as a backend for a gluster volume, it also becomes relevant for gluster.

Do we maintain this separation? If yes, do we simply ensure that we link off from the gluster namespace to the appropriate node namespace to map volumes and bricks to block devices? What is the impact of this approach?

The approach in this spec seems to be to gather the underlying device information in gluster's state sync rather than enhancing the node agent's inventory sync, which I would assume, would be quite beneficial to enable tendrl, irrespective of whether gluster is in the picture or not, to monitor the underlying block device.

nnDarshan:
https://github.com/Tendrl/specifications/pull/201#pullrequestreview-55456710
As part of this spec we are trying to map underlying block devices for a gluster brick (mount point). Each brick will map to one or more block devices, the name of block devices that are under the brick are stored as attribute of the brick `clusters/<cluster-id>/Bricks/all/<brick-path>/devices`. Each of the device in the list can be mapped to block devices under every node details, the path would be `nodes/<node-id>/LocalStorage/BlockDevices/all/<device_path>`

@brainfunked Have updated this info under problem description section of this spec

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,120 @@
...
+This spec would add a new rpm(python-blivet) depencency on gluster-integraion
+
+== Testing
+
+Make sure that the underlying device details are properly synced for different types of
+devices.
+
+== Documentation Impact
+
+None
```
cybernth, line 114: Which repo python-blivet is available? Is there a wiki update required

specs/collect_brick_device_details.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+=== Notifications/Monitoring impact
+
+While there is a brick state change event, when we send the alerts we should also
+include the brick device information in the message.
+
+=== Other end user impact
+
+User will be able to view the underlying physical device info of a brick.
```
cybernth, line 69: How? This is just a backend change right?

specs/collect_brick_device_details.adoc
nnDarshan: python-blivet is readily available in centos7 base channel. And this package will be added as a dependency for tendrl-gluster-integration, so it will be automatically pulled and no wiki update needed

specs/collect_brick_device_details.adoc
nnDarshan: Yes, this spec is only w.r.t backed. The device info will be saved as attribute under brick object in etcd. Which in turn will enable API to give that info.

specs/collect_brick_device_details.adoc
shtripat: You should add `tendrl-ansible` dependencies section for this additional package.

nnDarshan: python-blivet is a dependency for gluster-integration package with this change, hence I have added python-blivet as a package dependency for gluster-integration in gluster-integration spec. So from deployment perspective there is no explicit need to install python-blivet. It gets pulled automatically while installing gluster-integration
Please let me know if I am missing something

shtripat: Ack. In case its implicit dependency for `gluster-integration` we wont need `tendrl-ansible` explicit dependency I feel.

cybernth (approved):

Issue #188, opened 2017-07-19
Gluster metrics.


cloudbehl:
Which gluster metrics are to be enabled.
How each of the metrics is going to be implemented.
How each of the metrics is to be visualised in grafana.
The alert thresholds (defaults) in grafana for each metric.
Query requirements on the grafana side to enable individual metrics or their aggregation.
Any flows to be added to the node agent for configuring any collectd plugins needed by the individual metrics.

anivargi:
@anmolbabu  please associate the spec to the issue.

anmolbabu:
https://github.com/Tendrl/specifications/issues/213#issuecomment-318001791

PR #206, opened 2017-07-24
Gluster metrics

anmolbabu:
Gluster metrics

tendrl-bug-id: Tendrl/specifications#188

### DIFF
--- a/specs/gluster_metrics-milestone1.adoc
+++ b/specs/gluster_metrics-milestone1.adoc
@@ -1,0 +1,276 @@
+// vim: tw=79
+
+= Gluster Metrics - Milestone 1
+
+== Introduction
+
+This specification deals with monitoring tendrl managed gluster cluster.
+It explains the following:
+
+* Which gluster metrics are to be enabled.
+* How each of the metrics is going to be implemented.
+* Any flows to be added to the node agent for configuring any collectd plugins
+  needed by the individual metrics.
+This specification will deal with metrics implemented in milestone 1
+
+== Problem description
+
+Admin using tendrl for gluster cluster management, should be able to see an
+easy to interpret representation of his gluster cluster/s with appropriate
+utilization trend patterns and any issue in his gluster cluster/s should be
+with no/minimal effort be made noticeable to the admin.
+
+== Use Cases
+
+A gluster cluster is imported(or created) in tendrl. From this point, this
+cluster needs to be monitored for utilization and status behaviour.
+
+== Proposed change
+
+The gluster metrics will be enabled at the very level from collectd and any
+aggregations thereon will be done by the grafana either on its own or using the
+aggregation rules as communicated to it by the monitoring-integration module.
+
+* Following are the gluster metrics to be enabled from collectd along with the
+  source of the information:
+    ** Brick level:
+        *** IOPS -- gluster volume profile <vol_name> info --xml
+
+            xml attribute hierarchy in cli response:
+            cliOutput -> volProfile -> brick -> cumulativeStats -> totalRead
+            cliOutput -> volProfile -> brick -> cumulativeStats -> totalWrite
+            iops <- totalRead + totalWrite
+
+        *** Brick Utilization -- python os.statvfs on brick path
+
+            total -> data.f_blocks * data.f_bsize
+            free -> data.f_bfree * data.f_bsize
+            used_percent -> 100 - (100.0 * free / total)
+
+        *** Inode utilization -- python os.statvfs on brick path
+
+            free -> data.f_ffree
+            total -> data.f_files
+            used_percent_inode -> 100 - (100.0 * free_inode / total_inode)
+
+        *** LVM Thin Pool metadata usage -- lvm vgs command
+        *** LVM thin pool data usage -- lvm vgs command
+
+        ----
+        l = map(lambda x: dict(x),
+                map(lambda x: [e.split('=') for e in x],
+                    map(lambda x: x.strip().split('$'), out)))
+        d = {}
+        for i in l:
+            if i['LVM2_LV_ATTR'][0] == 't':
+                k = "%s/%s" % (i['LVM2_VG_NAME'], i['LVM2_LV_NAME'])
+            else:
+                k = os.path.realpath(i['LVM2_LV_PATH'])
+            d.update({k: i})
+        return d
+
+        if lvs and device in lvs and \
+           lvs[device]['LVM2_LV_ATTR'][0] == 'V':
+            thinpool = "%s/%s" % (lvs[device]['LVM2_VG_NAME'],
+                                  lvs[device]['LVM2_POOL_LV'])
+            out['thinpool_size'] = float(
+                lvs[thinpool]['LVM2_LV_SIZE']) / 1024
+            out['thinpool_used_percent'] = float(
+                lvs[thinpool]['LVM2_DATA_PERCENT'])
+            out['metadata_size'] = float(
+                lvs[thinpool]['LVM2_LV_METADATA_SIZE']) / 1024
+            out['metadata_used_percent'] = float(
+                lvs[thinpool]['LVM2_METADATA_PERCENT'])
+            out['thinpool_free'] = out['thinpool_size'] * (
+                1 - out['thinpool_used_percent'] / 100.0)
+            out['thinpool_used'] = out['thinpool_size'] - out['thinpool_free']
+            out['metadata_free'] = out['metadata_size'] * (
+                1 - out['metadata_used_percent'] / 100.0)
+            out['metadata_used'] = out['metadata_size'] - out['metadata_free']
+        ----
+
+    ** Node level:
+        *** CPU -- collectd plugin
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
+        *** SWAP -- collectd plugin
+        *** Disk IOPS -- collectd plugin
+        *** Network
+            **** Bytes sent and received(rx, tx) -- collectd plugin
+            **** Dropped packets -- collectd plugin
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
+        *** Status -- gluster get-state glusterd odir /var/run file collectd-glusterd-state
+            **** Counters will be made available via collectd plugins based on get-state result
+        *** Bricks counter
+            **** Total -- gluster get-state glusterd...
+    ** Cluster level:
+        *** Number of nodes
+            **** Total -- gluster get-state glusterd...
+            **** Status-wise
+                 Based on "peer in cluster" and "connected" in get-state output
+        *** Number of volumes
+            **** Total -- gluster get-state glusterd...
+            **** Status wise -- gluster get-state glusterd...
+        *** Number of bricks
+            **** Total -- gluster get-state glusterd...
+    ** Brick and cluster status wise counters can't be made  available from the
+       level of collectd as only the brick info corresponding to node on which
+       the collectd plugin is currently in execution, will be accessible and
+       info regarding the other bricks in cluster is not available and plugins
+       are stateless. This if required will not be found and updated to
+       graphite by monitoring-integration.
+* Collectd plugins for each of the above mentioned metrics will be added and
+  configured as detailed in :
+
+----
+https://github.com/Tendrl/specifications/blob/master/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc#proposed-change
+----
+
+----
+Note:
+The aggregated/derived metrics that will be made available at grafana
+level will be dealt with in detail by spec issues:
+    * https://github.com/Tendrl/specifications/issues/179
+----
+
+=== Alternatives
+
+* A new pluggable approach within our gluster metric collectors can be
+  developed such that the metrics collection plugins are loaded and
+  activated dynamically given just a one time configuration specifying
+  cluster_id and graphite host and port addresses.The plugins hooked to
+  gluster metrics provided pluggable framework are executed in concurrent as
+  green threads(being experimented) and hence their atomicity needs to be
+  ensured.
+    * The cluster topology is learnt once per execution cycle and the same
+      knowledge is shared between plugins unless the same can be fetched
+      by using a different means which is unavoidable.
+      ex: gluster volume status clients info --xml should be used to get
+      connections count but that already readily gives cluster topology
+      which can be used directly.
+    Note:
+        * Above approach runs every plugin in the framework at same interval.
+        * Attempts are being made to have a mechanism by which plugins can
+          register to our framework as light-weight, heavy-weight etc.. and
+          accordingly, the plugin execution interval will differ.
+        Note:
+        ** This is of particular importance in case of certain operations where
+           the command execution too frequently can affect the performance.
+* This is achieved as follows:
+    ** The plugins are classified as heavy-weight(can't be executed frequently
+       due to performance concerns) and low-weight(can be executed frequently)
+       under the collectd plugin path(/usr/lib64/collectd/gluster)
+    ** The base plugin executes gluster get-state command and learns cluster
+       topology.
+    ** The plugin base framework loads and executes as greenlets; all plugins
+       under above mentioned directories dynamically at every fixed interval
+       of time as configured in base plugin and passes cluster topology to
+       each of the plugins.
+
+----
+Note: This proposed alternative is being experimented.
+----
+
+=== Data model impact:
+
+The name-spacing of metrics will follow the following:
+
+* tendrl.clusters.<cluster_id> will be the prefix for all metrics.
+* Node level metrics follow the format:
+   tendrl.clusters.<cluster_id>.nodes.<node_name>.<plugin_name>.<plugin_attrs>
+* Cluster level metrics follow the format:
+    tendrl.clusters.<cluster_id>.<plugin_name>.<plugin_attrs...>
+* Volume level metrics follow the format:
+    tendrl.clusters.<cluster_id>.volumes.<volume_name>.<plugin_name>.
+    <plugin_attrs>
+* Brick level metrics follow the format:
+    tendrl.clusters.<cluster_id>.volumes.<volume_name>.nodes.<node_name>.bricks.
+        <brick_path>.<plugin_name>.<plugin_attrs>
+    and the same would also be maintained @
+    tendrl.clusters.<cluster_id>.nodes.<node_name>.bricks.<brick_path>.
+        <plugin_name>.<plugin_attrs>
+    for mapping directly from node level.
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+None
+
+==== Notifications/Monitoring impact:
+
+The configuration method would now change slightly in accordance with
+details in https://github.com/Tendrl/specifications/blob/master/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc#proposed-change
+
+==== Tendrl/common impact:
+None
+
+==== Tendrl/node_agent impact:
+
+None
+
+==== Sds integration impact:
+
+None
+
+=== Security impact:
+
+None
+
+=== Other end user impact:
+
+The main consumer of this is the tendrl-grafana dashboard.
+The impact in and due to the new dashboard will be detailed in a different
+spec.
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+None
+
+=== Developer impact:
+
+* The aggreageations that were previously done by the performance-monitoring
+  application will now be done at grafana.
+    ** The rules of aggregation will be communicated to grafana by
+       monitoring-integration. Details of this are out of scope of this spec
+       and will be covered as part of:
+
+       ----
+       https://github.com/Tendrl/specifications/pull/218
+       ----
+
+== Implementation:
+
+
+=== Assignee(s):
+
+Primary assignee:
+  Metrics from collectd: anmolbabu<anmolbudugutta@gmail.com>
+  Dashboard: cloudbehl<cloudbehl@gmail.com>
+
+=== Work Items:
+
+* https://github.com/Tendrl/node-agent/issues/549
+
+== Dependencies:
+
+None
+
+== Testing:
+
+The plugins push stats to graphite and the same in graphite will need to be
+tested for correctness.
+
+== Documentation impact:
+
+None
+
+== References:
+
+Attempts in this regard can be found @:
+https://github.com/Tendrl/node-monitoring/pull/79


specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,242 @@
...
+* Which gluster metrics are to be enabled.
+* How each of the metrics is going to be implemented.
+* The alert thresholds (defaults) in grafana for each metric.
+* Any flows to be added to the node agent for configuring any collectd plugins
+  needed by the individual metrics.
+
+
+== Problem description
+
+Admin using tendrl for gluster cluster management, should be able to see the
```
shtripat: s/see the a easy/see an easy/

anmolbabu:
This is WIP spec with some unknowns:
status wise counters for bricks:
This brick status is availble in get-state for the brick belonging to the node on which the get-state command is fired.. So there's no way collectd residing on a node can push this counter. The best that can be done from collectd's perspective is to push brick status as a integral number(as ceph-metrics does) to graphite and a way to aggregate this at grafana needs to be found..
Note This is not a mere count query. As the actual status is number encoded in the metric value and not in metric name.. Encoding it in metric name makes no sense as when the status changes, the previous metric and the current metric both co-exist leading to unnecessary confusions

@cloudbehl Please add the grafana side of things either to the same spec or to a different one..

@r0h4n @brainfunked @shtripat @nthomas-redhat @anivargi  Please review

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,242 @@
...
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
+        *** SWAP -- collectd plugin
+        *** Disk IOPS -- collectd plugin
+        *** Network
+            **** Bytes sent and received(rx, tx) -- collectd plugin
+            **** Dropped packets -- collectd plugin
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
+        *** Status -- gluster get-state glusterd odir /var/run file glusterd-state
```
shtripat: When you say `get-state` do you mean sds sycn from `gluster-integration` or collectd as well runs this every few seconds?

anmolbabu: Collectd also runs this. So I believe its better to use a different file for get-state o/p something like:
gluster get-state glusterd odir /var/run file collectd-glusterd-state
So that the files read and deleted by gluster-integration and collectd will differ and they don't end-up in a race condition

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,242 @@
...
+  gluster:
+    cluster_utilization:
+      Warning: 75
+      Failure: 90
+    volume_utilization:
+      Warning: 75
+      Failure: 90
+    brick_utilization:
+      Warning: 75
+      Failure: 90
```
cloudbehl: @rishubhjain @GowthamShanmugam I think these threshold values you will need to while implementing grafana's alerting.

anivargi: Yeah, the thresholds will be part of Grafana configuration.

anmolbabu:
https://github.com/Tendrl/specifications/issues/213#issuecomment-318001791

shtripat (commented):
LGTM. No specific comments.
@Tendrl/tendrl-core plz review.

r0h4n (dismissed):

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+aggregation rules as communicated to it by the monitoring-integration module.
+
+* Following are the gluster metrics to be enabled from collectd along with the
+  source of the information:
+    ** Brick level:
+        *** IOPS -- gluster volume profile <vol_name> info --xml
+
+            xml attribute hierarchy in cli response:
+            cliOutput -> volProfile -> brick -> cumulativeStats -> totalRead
+            cliOutput -> volProfile -> brick -> cumulativeStats -> totalWrite
```
cybernth: I understand that read and write collected but please mention how IOPS is calculated

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+
```
cybernth: How the drill down from brick to disk is addressed? Are you planning to collect the disk level iostat or at a mount point level? If the brick is created on RAID, then how this is addressed? please provide the details here

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+      which can be used directly.
+    Note:
+        * Above approach runs every plugin in the framework at same interval.
+        * Attempts are being made to have a mechanism by which plugins can
+          register to our framework as light-weight, heavy-weight etc.. and
+          accordingly, the plugin execution interval will differ.
+        Note:
+        ** This is of particular importance in case of certain operations where
+           the command execution too frequently can affect the performance.
+* A generic flow as in:
```
cybernth: You can link https://github.com/Tendrl/specifications/pull/219 here

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+The configuration of collectd from monitoring-integration will slightly change
+in terms of number of plugins to configure and the attributes to be passed in
+configuration.
+
+=== Developer impact:
+
+* The aggreageations that were previously done by the performance-monitoring
+  application will now be done at grafana.
+    ** The rules of aggregation will be communicated to grafana by
+       monitoring-integration. Details of this are out of scope of this spec
```
cybernth: Can you link to as spec/issue if that exists?

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+        *** Number of connections -- gluster volume status all clients --xml
+    ** Cluster level:
+        *** Quorum status
+        *** Number of nodes
+            **** Total -- gluster get-state glusterd...
+            **** Status-wise
+                 Based on "peer in cluster" and "connected" in get-state output
+        *** Number of volumes
+            **** Total -- gluster get-state glusterd...
+            **** Status wise -- gluster get-state glusterd...
```
cybernth: near full?

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+        *** Number of bricks
+            **** Total -- gluster get-state glusterd...
+    ** Brick and cluster status wise counters can't be made  available from the
+       level of collectd as only the brick info corresponding to node on which
+       the collectd plugin is currently in execution, will be accessible and
+       info regarding the other bricks in cluster is not available and plugins
+       are stateless. This if required will not be found and updated to
+       graphite by monitoring-integration.
```
cybernth: These are required, can grafana do this?

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
+        *** Status -- gluster get-state glusterd odir /var/run file collectd-glusterd-state
+            **** Counters will be made available via collectd plugins based on get-state result
+        *** Bricks counter
+            **** Total -- gluster get-state glusterd...
+        *** Pending Heal -- gluster volume heal all statistics
+        *** Rebalance status -- gluster get-state glusterd..
+        *** Number of connections -- gluster volume status all clients --xml
+    ** Cluster level:
```
cybernth: Please mention how the these metrics are collected/calculated(grafana?)
1. Cluster utilization
2. cluster iops, latency
3. Disks stats at cluster level(aggregate - iops, latency, throghput, nearful etcs)
4.Network stats
5. Hosts with high memory and cpu utilization

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+        *** CPU -- collectd plugin
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
+        *** SWAP -- collectd plugin
+        *** Disk IOPS -- collectd plugin
+        *** Network
+            **** Bytes sent and received(rx, tx) -- collectd plugin
+            **** Dropped packets -- collectd plugin
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
```
cybernth: utilization?
iops/latency
Disks stats at cluster level(aggregate - iops, latency, throghput, nearful etcs)

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+        *** Status -- gluster get-state glusterd odir /var/run file collectd-glusterd-state
+            **** Counters will be made available via collectd plugins based on get-state result
+        *** Bricks counter
+            **** Total -- gluster get-state glusterd...
```
cybernth: brick status wise counts?

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,272 @@
...
+                1 - out['metadata_used_percent'] / 100.0)
+            out['metadata_used'] = out['metadata_size'] - out['metadata_free']
+        ----
+
+    ** Node level:
+        *** CPU -- collectd plugin
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
+        *** SWAP -- collectd plugin
+        *** Disk IOPS -- collectd plugin
```
cybernth: Is this individual disk level or aggregate for node like  iops, latency, throghput, nearful etc
Also are we planning to collect brick status wise counts here?

cybernth (changes requested):
I understand that lot of aggregation is done at grafana level. But we need capture them here and clearly state that it will be handled at grafana

specs/gluster_metrics.adoc
anmolbabu: Drill down is something we need to figure out yet and it is slated for implementation in subsequent drops.. It is most likely that the relation from brick to its disk needs to be communicated to grafana by monitoring-integration.The details of how this happens is not yet found out @anivargi can you please add if I am missing something
As of now what we have is :
- Disk Plugin(https://collectd.org/wiki/index.php/Plugin:Disk) that gives performance statistics(iops and io time) of hard-disks and, where supported, partitions.
- df plugin(https://collectd.org/wiki/index.php/Plugin:DF) that gives file system usage information, i. e. basically how much space on a mounted partition is used and how much is available.
The iostat plugin is not packaged with collectd I will need to read more in this regard..

Also, the spec only currently deals with whatever is available today and for the upcoming immediate drop.. The spec is expected to evolve as we progress..

specs/gluster_metrics.adoc
anmolbabu: Aggregations are planned to be done by grafana in accordance with rules configured by monitoring-integration.. @anivargi or @cloudbehl can you add details in this regard like how those rules can be comminicated..

All these are not slated for upcoming drop and hence the spec will evolve to contain these information..

specs/gluster_metrics.adoc
anmolbabu: Aggregations are planned to be done by grafana in accordance with rules configured by monitoring-integration.. @anivargi or @cloudbehl can you add details in this regard like how those rules can be comminicated..

All these are not slated for upcoming drop and hence the spec will evolve to contain these information..

specs/gluster_metrics.adoc
anmolbabu: Collectd plugin does not have access to statuses of bricks from other nodes(get-state provide brick status of brick from the node where it is currently executed). So, monitoring-integration which will have access to etcd(where gluster-integrations from all gluster nodes sync their respective brick statuses to etcd) can push this stat to graphite as an exception

@nthomas-redhat @anivargi @brainfunked  Please suggest

specs/gluster_metrics.adoc
anmolbabu: Aggregations are planned to be done by grafana in accordance with rules configured by monitoring-integration.. @anivargi or @cloudbehl can you add details in this regard like how those rules can be comminicated..

All these are not slated for upcoming drop and hence the spec will evolve to contain these information..

Regarding hosts that cross threshold kind of information @cloudbehl since the thresholds are configured at grafana end, is this feasible.
Also get top 5 most used kind of thing is that possible in grafana?? @cloudbehl Cc @anivargi

specs/gluster_metrics.adoc
anmolbabu: Regarding metrics or entities that cross threshold kind of information @cloudbehl since the thresholds are configured at grafana end, is this feasible via grafana.
Also get top 5 most used kind of thing is that possible in grafana?? @cloudbehl Cc @anivargi

specs/gluster_metrics.adoc
anmolbabu: Nope. We need some other module(like monitoring-integration) to be able to do this

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
shtripat: I agree. As part of perfirmance improvements, the similar thing is planned for utilization, status, connections aggregations in gluster sds sync as well. IN that case we would have a separate thread which executes a flow once cluster is imported/created to sync the aggregated details and schedules it to run at regular intervals post that.
Something similar would be required here as well I feel.

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,277 @@
...
+            out['thinpool_used'] = out['thinpool_size'] - out['thinpool_free']
+            out['metadata_free'] = out['metadata_size'] * (
+                1 - out['metadata_used_percent'] / 100.0)
+            out['metadata_used'] = out['metadata_size'] - out['metadata_free']
+        ----
+
+    ** Node level:
+        *** CPU -- collectd plugin
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
```
nnDarshan: We should probably decide on set of predefined mounts that we need to monitor right ? like /root, /boot, /home.

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,277 @@
...
+            **** Dropped packets -- collectd plugin
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
+        *** Status -- gluster get-state glusterd odir /var/run file collectd-glusterd-state
+            **** Counters will be made available via collectd plugins based on get-state result
+        *** Bricks counter
+            **** Total -- gluster get-state glusterd...
+        *** Pending Heal -- gluster volume heal all statistics
+        *** Rebalance status -- gluster get-state glusterd..
+        *** Number of connections -- gluster volume status all clients --xml
```
nnDarshan: I guess this is a costly operation. However get-state gives you per brick client info that is present in that particular node. I feel that can be used, what do you think ?

specs/gluster_metrics.adoc
```diff
@@ -0,0 +1,277 @@
...
+        *** CPU -- collectd plugin
+        *** Memory -- collectd plugin
+        *** Storage(Mount point) -- collectd plugin
+        *** SWAP -- collectd plugin
+        *** Disk IOPS -- collectd plugin
+        *** Network
+            **** Bytes sent and received(rx, tx) -- collectd plugin
+            **** Dropped packets -- collectd plugin
+            **** Errors/overruns -- collectd plugin
+    ** Volume level:
```
nnDarshan: Hey, isn't snapshot count for a volume, number of geo-rep sessions for a volume (with ok and faulty count) needed

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Why??

I feel we can expose all data but allow UI to decide what it wants to show.. And if we limit the mount-points to be monitored, there is every chance we loose out some or many important mount points created from cli(ex: volume mounts, disk mounts etc..)..

Otherwise the moment we detect even some of these mount points via any tendrl service the collectd config file responsible for mount point stats will need to reconfigured and collectd needs to be restarted every time on any(all) nodes where ever this happens..

specs/gluster_metrics.adoc
anmolbabu: I don't see this currently from get-state o/p is it planned for some-time soon??

nnDarshan: This is available in latest version(this was added recently). you can check gluster nightly builds

specs/gluster_metrics.adoc
anmolbabu: Hmm I think we narrowed the scope to https://docs.google.com/spreadsheets/d/1X2JREn0TybSqLHgx9vcv-zjn4S9X9_Xz0JdhADew5-Q/edit?ts=5965bec8#gid=1819780204 and I missed this but that shouldn't be too difficult to add from get-state o/p.
I'll add them..

specs/gluster_metrics.adoc
cybernth: Agreed, but this spec should list all the metrics and specifically state that this will done at grafana. Also link with spec if any

specs/gluster_metrics.adoc
cybernth: This spec should list all the metrics and specifically state that this will done at grafana. Also link with spec if any

specs/gluster_metrics.adoc
nnDarshan: only concern with monitoring all mount points on the node is that all the glusterfs bricks are mount points and we are monitoring them as part of brick utilization collection. if we monitor all mount points we will be repeating the monitoring of gluster brick mount points.

specs/gluster_metrics.adoc
cybernth: @anmolbabu , what is discussed earlier was to collect individual status per node and aggregate at grafana right?

specs/gluster_metrics.adoc
cybernth: This spec should list all the metrics and specifically state that this will done at grafana. Also link with spec if any

specs/gluster_metrics.adoc
cybernth: this spec should list all the metrics and specifically state that this will done at grafana. Also link with spec if any

specs/gluster_metrics.adoc
cybernth: Why grafana can't do this?

specs/gluster_metrics.adoc
anmolbabu: I acknowledge this.. But then if monitoring-integration can bury this info in templates that it communicates to grafana, I would prefer that rather than limiting collectd to monitor only a pre-defined set of mount points.. The reason is as I said, if we ever dynamically discover that we need a new mount-point to be monitored, that involves the following:
1. collectd config change
2. restart collectd to reflect config change to it.
and this on every node where the change happened..

specs/gluster_metrics.adoc
anmolbabu: @nthomas-redhat Yes that stands to be the case even now. 
Only thing is the aggregation logic like for example, brick utilization to volume utilization aggregation should not ideally be a mere summation of utilization of all bricks but instead this differs based on volume type.. So these agregation logic need to be communicated to grafana from monitoring-integration..

specs/gluster_metrics.adoc
anmolbabu: Because the individual brick statuses are mapped to integeral values and grafana I am not sure if it can look at the value of metric and classify them based on values and count each of the classifications @cloudbehl is this facilitated by grafana??

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Done

specs/gluster_metrics.adoc
anmolbabu: Ack

specs/gluster_metrics.adoc
nnDarshan: ack.

nnDarshan:
Looks good to me

anmolbabu:
@r0h4n @nthomas-redhat @shtripat @brainfunked As discussedI have trimmed this spec to deal with milestone1 part
I'll raise Gluster Metrics-milestone 2 once this is merged
Please review..

specs/gluster_metrics-milestone1.adoc
```diff
@@ -0,0 +1,297 @@
...
+* A generic flow as in:
+
+----
+https://github.com/Tendrl/node-monitoring/blob/master/tendrl/node_monitoring/flows/configure_collectd/__init__.py
+----
+
+and a genric command to generate config file from template as in:
+
+----
+https://github.com/Tendrl/node-monitoring/blob/master/tendrl/node_monitoring/commands/config_manager.py
```
shtripat: Change the name of the binary as its `tendrl_monitoring_config_manager.py`

specs/gluster_metrics-milestone1.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+None
+
+==== Notifications/Monitoring impact:
+
+The configuration method would now change slightly in accordance with
```
shtripat, line 203: Can we add more details here or `Proposed Change` section has enough details?

specs/gluster_metrics-milestone1.adoc
```diff
@@ -0,0 +1,297 @@
...
+The impact in and due to the new dashboard will be detailed in a different
+spec.
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+The configuration of collectd from monitoring-integration will slightly change
```
shtripat: Same as above because `The configuration of collectd from monitoring-integration will slightly change` creates confusion I feel.

specs/gluster_metrics-milestone1.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+=== Assignee(s):
+
+Primary assignee:
+  Metrics from collectd: anmolbabu<anmolbudugutta@gmail.com>
+  Dashboard: cloudbehl<cloudbehl@gmail.com>
+
+=== Work Items:
+
+Github issues will be raised and linked here once the repo that hosts the
```
shtripat: I think this is available now as `monitoring-integration`. Can we add specific issues here?

specs/gluster_metrics-milestone1.adoc
anmolbabu: Removing them from here they are already briefed in https://github.com/Tendrl/specifications/blob/master/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc#proposed-change

specs/gluster_metrics-milestone1.adoc
anmolbabu: Removing them from here they are already briefed in https://github.com/Tendrl/specifications/blob/master/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc#proposed-change

specs/gluster_metrics-milestone1.adoc
anmolbabu: yes removing it as this configuration happens internally and requires no user intervention

specs/gluster_metrics-milestone1.adoc
anmolbabu: Done I just needed to link issue on node-agent https://github.com/Tendrl/node-agent/issues/549

specs/gluster_metrics-milestone1.adoc
shtripat: Ack

specs/gluster_metrics-milestone1.adoc
shtripat: Ack. Can we have some details of these internal configurations.

specs/gluster_metrics-milestone1.adoc
```diff
@@ -0,0 +1,276 @@
...
+
+----
+https://github.com/Tendrl/specifications/blob/master/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc#proposed-change
+----
+
+----
+Note:
+The aggregated/derived metrics that will be made available at grafana
+level will be dealt with in detail by spec issues:
+    * https://github.com/Tendrl/specifications/issues/179
```
rishubhjain, line 134: If possible could you provide a link instead of text?

Issue #180, opened 2017-07-19
Enable Core gluster alerts from tendrl.


nnDarshan:
Event generation for alert condition setting from gluster-integration. Gluster eventing framework and state sync both contribute to these events being generated in tendrl. gluster-integration would generate and process the events and hand them over to the node agent for transport and alert condition setting.

This spec requires a list of specific gluster events that should be treated as events in the tendrl alert lifecycle.

PR #207, opened 2017-07-24
Enable glusterfs native events in tendrl.

nnDarshan:
tendrl-bug-id: Tendrl/specifications#180
Signed-off-by: nnDarshan <darshan.n.2024@gmail.com>


nnDarshan:
@aravindavk @shtripat @Tendrl/tendrl-core Please review

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+
+Currently tendrl is not utilizing the events exposed by glusterfs events API.
+Tendrl should be enhanced to consume and handle the events.
+
+== Use Cases
+
+* Glusterfs events API pushes events for certain internal glusterfs failure conditions
+which is not track able by tendrl. Tendrl can raise suitable alerts for such events.
+
+* The interval of sds state sync can be increased (to 30 min) and we should have event
```
shtripat, line 16: I feel 30 min is just an example and still it would configurable ;)

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+== Proposed change
+
+The event receiver web-hook can be run on all the gluster storage nodes as a
+thread/greenlet in gluster-integration daemon. And we should configure the
+gluster eventing daemon to push events in localhost mode (the events generated
+by gluster eventing daemon will be pushed to the web-hook running only on
+that particular node). With this we are eliminating single point of failure
+w.r.t event listener.
+
+Since some of the events are raised from all the nodes or raised multiple times
```
shtripat, line 29: Do you mean here that all the nodes would first update the events in central store as is. Later there would be a dedup logic to make meaning out of that? Can you elaborate more?

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+have a valid state in case events are missed.
+
+=== Alternatives
+
+An alternate approach to listen events is to have listener enabled in only one of
+the gluster-integration process, and configure all gluster events API to push
+events to this instance. Though this has slight advantage of not having to run the
+event listeners on all the nodes.
+
+The main problem with this approach is handling the failure of this listener node.
```
shtripat, line 45: I feel this problem is like `monitoring-integration` where we plan to do certain aggregations work. We ideally should be following similar approach which `monitoring-integration` does in future.

Another wild thought, if we can have this `webhook` maintained as part of `monitoring-integration` component itself and do the required dedup etc in this component. @Tendrl/tendrl-core comments?

The problem with this model is though, you would loose out on the thought of running `sds_sync` of `gluster-integration` only on an event.

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+to the native events. Performance evaluation has to be made after adding listener to
+gluster integration process.
+
+On the other hand, if we increase the gluster integration sds sync interval to (say
+30 mins) and have the sync driven by events. There will be a considerable improvement
+in the performance.
+
+=== Other deployer impact
+
+We would need to install an additional package glusterfs-events on all the storage
```
shtripat, line 85: Isnt this daemon started and running by default with glusterfs?

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+
+* Get complete list of all supported events from glusterfs team
+
+* Get a list of alerts among all events from glusterfs team, these alerts
+will be used in phase 1
+
+== Testing
+
+* Make sure that there is no great performance degradation after implementing this spec,
+as this spec involves addition of a new thread to gluster-integration process
```
shtripat, line 135: Another service as well :)

specs/enable_gluster_native_events.adoc
nnDarshan: yes, it was an example and we can make it configurable :)

specs/enable_gluster_native_events.adoc
nnDarshan: No we update the event to etcd, only if the state that is received from event is different from the state in etcd and we update the latest state to etcd. Thus in case of multiple duplicate events only the first one is saved in etcd

specs/enable_gluster_native_events.adoc
nnDarshan: Nope

specs/enable_gluster_native_events.adoc
shtripat: ack

specs/enable_gluster_native_events.adoc
nnDarshan: Few drawbacks with this approach:
* It might not scale with multiple clusters as all the clusters will push notifications to this node and handling might be difficult.
* It is better to keep the event handling code close to gluster integration as it needs to access lot of objects in gluster namespace
* monitoring integration service becomes a single point of failure, If some thing goes wrong in the service, events from all the clusters are lost

shtripat: Ack. I feel based on response for previous comment, this looks fine to me. If we check the state value for the affected entity in central store and then to take action on the event, its fine.

Lets look for comments from others.

specs/enable_gluster_native_events.adoc
```diff
@@ -0,0 +1,147 @@
...
+that particular node). With this we are eliminating single point of failure
+w.r.t event listener.
+
+Since some of the events are raised from all the nodes or raised multiple times
+(till alert condition gets fixed), tendrl should make sure that duplicate events
+are handled appropriately by checking the state information stored in etcd.
+
+In future, gluster integration state sync should be based on events received
+by glusterfs events API. Thus we make sure that we do the state sync only when
+there is a state change. Also we can have a 30 min sync, to make sure that we
```
brainfunked, line 35: 30 minutes is rather long. I'd stick to not more than 5 minutes.

Also, would we need a locking mechanism to ensure that multiple state syncs don't run simultaneously. Is there any part of the gluster state sync that happens at a level where more than one gluster-integration instances would be involved?

specs/enable_gluster_native_events.adoc
brainfunked: My earlier comment asks about whether any part of the gluster-integration state sync, as currently implemented, is NOT node specific. If the state sync for each of the integration instances is independent of the others, it's better to go with the original approach rather than this alternative one.

brainfunked (commented):
I think my involvement in this spec would be limited to only the architectural concerns. As such, unless there's an explicit NACK from me, please feel free to merge the spec @r0h4n @nthomas-redhat .

specs/enable_gluster_native_events.adoc
nnDarshan: I don't see any issues if state sync's happen simultaneously. Each gluster-integration will update the state of cluster as it sees from that node.
@shtripat what do you feel

shtripat: So at the moment the aggregations at cluster level like cluster  utilization, cluster health status, cluster client connections etc are done at last post volume sync. There are plans to move this aggregation logic as a flow which would be triggered while starting the `{gluster/ceph}_integration` and after first successful run, this job would self schedule at regular interval. This would be picked by one node of the cluster randomly and processed.
Hope this makes clear.

Issue #182, opened 2017-07-19
Initial UI refresh for the grafana integration


a2batic:
Cluster specific objects like volumes and pools are to be removed from the main navigation bar. These views would only be available under a specific cluster. Update the navigation to allow quick switching between different clusters.

brainfunked:
The following navigation entries would be available post the refresh from the [ui-legacy-code](https://github.com/Tendrl/ui/tree/ui-legacy-code) branch, which was created as part of the codebase refresh addressed in https://github.com/Tendrl/ui/issues/486.

* Remove the global navigation items for cluster specific objects such as volumes, pools, bricks etc.
* Retain the global navigation entries for Clusters, Hosts, Admin.
* Retain the global navigation entry for Alerts.
* Retain the top navigation entry for the notifications panel and authentication.

brainfunked:
The following functionality would be carried forward from the [ui-legacy-code](https://github.com/Tendrl/ui/tree/ui-legacy-code) branch, which was created as part of the codebase refresh addressed in https://github.com/Tendrl/ui/issues/486.

* Notifications drawer
* Global alerts list
* Hosts list
* Tasks list
* Task details page
* Authentication

Other functionality such as workflows, list views and the various object details views would be ported subsequently. The updated cluster list views are being addressed in spec https://github.com/Tendrl/specifications/issues/181, while the import flow would be ported in spec https://github.com/Tendrl/specifications/issues/183.

ltrilety:
I have several doubts about this issue. As the issue doesn't correspond to what was said in https://github.com/Tendrl/specifications/issues/200

Especially these two points makes me wonder:

- Remove the global navigation items for cluster specific objects such as volumes, pools, bricks etc.
- Retain the global navigation entries for Clusters, Hosts, Admin.
<br>
As I see it:

 - Via [Design A](https://redhat.invisionapp.com/share/2RCOE72PE) there should be `Clusters`, `Users`, `Mail Settings` and `SNMP Settings` navigation items plus upper menu with `notifications` no `Hosts` no `Admin` etc.
 - Via [Design B](https://redhat.invisionapp.com/share/8QCOEVEY9) there should be no just context selector and upper menu for users, snmp plus smtp settings, and alerts once again no `Hosts` etc.
 - Via mockups provided by @gnehapk there should be all navigation items including `volumes`, `bricks` etc.

So my question is what design is followed, if any?
[...]

PR #214, opened 2017-07-25
Specification for Global Navigation

a2batic:
Tendrl-bug-id: #182


brainfunked:
Can I assume that this one was closed in favour of https://github.com/Tendrl/specifications/pull/215 and directly review that one instead @a2batic?

Issue #203, opened 2017-07-24
Alert Lifecycle Design


brainfunked:
The lifecycle of an event from being generated to be notified to the administrator is a four stage process: event generation, event processing to add tendrl-relevant metadata, setting the alert condition and notifying the administrator of the alert condition change. Along with this, the node agent serves as the transport to set the alert condition in etcd.

PR #217, opened 2017-07-26
Alerting system design initial spec

brainfunked:
Still work in progress. Additional references need to be added. Potentially more details regarding the implementation guidelines for the various components involved may need to be added as well.

tendrl-bug-id: https://github.com/Tendrl/specifications/issues/203


GowthamShanmugam (approved):
This is solved all of my doubts in alert notification issue. It is looks good to me

brainfunked:
@anmolbabu could you please link here to the earlier spec(s) where the `priority` and `severity` attributes in a Message had been discussed? I need to update this spec based on that information.

anmolbabu:
@brainfunked https://github.com/Tendrl/specifications/blob/master/specs/pluggable_alert_delivery.adoc

GowthamShanmugam:
@brainfunked @anmolbabu @shtripat @nthomas-redhat @r0h4n 

Implementation steps of alerting architecture : 

1. When monitoring integration starts, at that time webhook will start and keep listen the grafana alerts.
2. When an alert raised by grafana the webhook will receive the alert and pass the alert into alerthandlersmanager
3. Using Rule id in grafana alert it will query grafana again to get full details about the particular alert.
4. Alert handler manager has different types of the alert handler to handle different alerts.  using the alert name in grafana alert it can identify particular handler.
        Node alert handlers for:
                1. cpu utilization
                2. memory utilization
                3. swap utilization
         Cluster alert handlers for:
                 1. cluster utilization
                 2. brick utilization
                 3. volume utilization
4. Based on alert name particular handler is called with grafana alert as param. And the handler will convert the grafana alert as tendrl alert and push that into message socket.  
5. Up to this level alert id is empty because we are going to use message id as alert id.
5. Message handler can identify the alert using priority of the message. ("notice").
6. if message handler identified alert message then it is passed to update the alert.
7. If the alert classification is node then the alert properties are compared with other alerts in node/node_id dir of etcd("alerting/nodes/{node_id}"), or if alert classification is cluster then the alert properties are compared with that particular cluster alerts in etcd (alerting/clusters/{cluster_id}). 
8. If alert properties match then it will assign the same alert_id for this new alert and update the old alert using new alert object using alert_id in etcd.
     * if alert properties are matched then a new alert is ok alert (in tendrl ok alert is INFO) then old alert is replaced by new clearing alert and ttl is set for 2 days. (if the alert no longer exist then it will be removed by ttl).
    * If alert properties are matched but the alert is not clearing alert (warning), then an old alert is replaced by new alert using alert id.  and ttl is removed. because it is warning alert.
    *  If alert properties are not matched with any alert then it will be treated as a new alert. so it will be updated using new alert id.

so the implementation is like if the alert property matches with some alert in etcd then same alert id is assinged to the new alert and updated  in etcd. if not matched then alert updated in etcd with new id.  This is decided after alert is received from socket.


When node-agent receive alert message from socket it will pass the alert in two places one is to the logger to log the alert in /messages/ in etcd. one more place is to compare alert properties and updates the alert in /alerting/  dir in etcd.

Issue #204, opened 2017-07-24
Refactoring of the node monitoring flows into node agent.


anmolbabu:
The present day node-monitoring contains the following:
- The collectd custom plugins(majority being sds-specific)
- Jinja2 templates for configuring the collectd plugins
- A generic flow and a python script(that can be used as a linux command) to get the configs generated from jinja templates given the appropriate parameters.

To be removed:
- A python script that pushes collectd detected threshold alert to etcd.

The idea behind a specification for this is to merge node-monitoring and node-agent.

Challenges In consolidating the node-monitoring and node-agent services:
- Packaging the sds-specific plugins and templates:
  If they are moved out to sds-integrations, then who takes the responsibility of running the generic configuration flow?? Will the flows for even sds configuration still be in node-agent and executed by node-agent itself??

anmolbabu:
Closing this as it is implemented by https://github.com/Tendrl/node-agent/pull/555

PR #219, opened 2017-07-27
Refactor node monitoring flows into node agent

anmolbabu:
tendrl-bug-id: Tendrl/specifications#204

### DIFF
--- a/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
+++ b/specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
@@ -1,0 +1,235 @@
+// vim: tw=79
+
+= Refactoring Of Node Monitoring Flows Into Node-Agent
+
+== Introduction
+
+This specification deals with detailing the work involved in merging the
+node-monitoring and node-agent services.
+
+Currently the responsibilities of node-monitoring are as follows:
+
+* It packages the collectd custom plugins
+    ** sds-specific utilization, status and counters plugins
+    ** node cluster network throughput plugin
+* Jinja2 templates for configuring the collectd plugins(both custom and
+  collectd provided)
+* A generic flow and a python script(that can be used as a linux command) to
+  get the configs generated from jinja templates given the appropriate parameters.
+
+== Problem description
+
+The idea behind having node-agent and node-monitoring seperated into a seperate
+packages was to facilitate modularity in tendrl stack such that users of tendrl
+stack can have the freedom to choose whether or not to have monitoring stack
+along with tendrl core stack. But, this pulls in overhead of having to maintain
+greater number of services.So, the idea is to now consolidate the services as
+much as possible and as part of this idea, this spec proposes consolidation of
+node-monitoring into node-agent.
+
+== Use Cases
+
+Admin creates/imports a cluster using/into tendrl and has monitoring enabled
+out of the box.
+
+== Proposed change
+
+* The generic flow as in:
+
+----
+https://github.com/Tendrl/node-monitoring/blob/master/tendrl/node_monitoring/flows/configure_collectd/__init__.py
+----
+
+and a genric command to generate config file from template as in:
+
+----
+https://github.com/Tendrl/node-monitoring/blob/master/tendrl/node_monitoring/commands/config_manager.py
+----
+
+will be moved to node-agent.
+
+* The templates currently in:
+
+----
+https://github.com/Tendrl/node-monitoring/tree/master/tendrl/node_monitoring/templates
+----
+
+and the plugins currently in:
+
+----
+https://github.com/Tendrl/node-monitoring/tree/master/tendrl/node_monitoring/plugins
+----
+
+will be moved into node-agent.
+
+* Earlier the jobs for initiating the flows to configure collectd on nodes were
+  initiated by performance-monitoring node. But for now(momentarily) as a final
+  step of import, create and expand cluster, this will be initiated on all
+  nodes of cluster.
+
+* Initiating(loading job to etcd) configuration on cluster nodes will involve
+  invoking configuration flow from an atom marked post atom in import cluster
+  flow.
+  ----
+  Job loading sample:
+  https://github.com/Tendrl/performance-monitoring/blob/master/tendrl/performance_monitoring/utils/util.py#L41
+  ----
+
+=== Alternatives
+
+None
+
+=== Data model impact:
+
+A new flow definition will be added to node-agent under its namespace as below:
+
+----
+namespace.node_agent:
+  flows:
+    ConfigureCollectd:
+      atoms:
+        - tendrl.objects.Node.atoms.Cmd
+      help: "Execute given command on given node"
+      enabled: true
+      inputs:
+        mandatory:
+          - Node.fqdn
+          - plugin_name
+          - plugin_conf_params
+          - Service.name
+      post_run:
+        - tendrl.objects.Service.atoms.CheckServiceStatus
+      pre_run:
+        - tendrl.objects.Node.atoms.CheckNodeUp
+      run: tendrl.node_agent.flows.ConfigureCollectd
+      type: Create
+      uuid: dc8fff3a-34d9-4786-9282-55eff6abb6c4
+      version: 1
+tendrl_schema_version: 0.3
+----
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+None
+
+==== Notifications/Monitoring impact:
+
+* node-monitoring project will be deprecated.
+* performance-monitoring will also be deprecated and details related to this are
+  captured as part of :
+    ** https://github.com/Tendrl/specifications/pull/218
+    ** https://github.com/Tendrl/specifications/pull/198
+
+==== Tendrl/common impact:
+None
+
+==== Tendrl/node_agent impact:
+
+As described in section "Proposed Change" section, collectd templates, custom
+plugins and flow to generate collectd configuration will move into node-agent.
+
+* The plugins and templates will go as under :
+
+----
+[node_agent]$ tree tendrl/node_agent/collectd/
+tendrl/node_agent/collectd/
+├── collectors
+│   ├── ceph
+│   │   ├── __init__.py
+│   │   ├── tendrl_ceph_cluster_iops.py
+│   │   ├── tendrl_ceph_node_network_throughput.py
+│   │   └── tendrl_ceph_utilization.py
+│   ├── glusterfs
+│   │   ├── heavy_weight
+│   │   │   ├── __init__.py
+│   │   │   └── tendrl_glusterfs_clients_info.py
+│   │   ├── __init__.py
+│   │   ├── low_weight
+│   │   │   ├── __init__.py
+│   │   │   ├── tendrl_glusterfs_brick_utilization.py
+│   │   │   ├── tendrl_glusterfs_health_counters.py
+│   │   │   ├── tendrl_glusterfs_peer_network_throughput.py
+│   │   │   └── tendrl_glusterfs_profile_info.py
+│   │   ├── tendrl_glusterfs_base.py
+│   │   └── utils.py
+│   ├── ini2json.py
+│   └── __init__.py
+├── __init__.py
+└── templates
+    ├── ceph
+    │   ├── tendrl_ceph_cluster_iops.jinja
+    │   ├── tendrl_ceph_cluster_utilization.jinja
+    │   └── tendrl_ceph_node_network_throughput.jinja
+    ├── glusterfs
+    │   ├── tendrl_glusterfs_clients_info.jinja
+    │   └── tendrl_glusterfs.jinja
+    └── node
+        ├── collectd.jinja
+        ├── cpu.jinja
+        ├── dbpush.jinja
+        ├── disk.jinja
+        ├── latency.jinja
+        ├── memory.jinja
+        ├── mount_point.jinja
+        ├── network.jinja
+        └── swap.jinja
+
+----
+
+==== Sds integration impact:
+
+None
+
+=== Security impact:
+
+None
+
+=== Other end user impact:
+
+The end user will now have only node-agent to be deployed on nodes apart from
+sds-integrations.
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+Tendrl ansible will now need to remove node-monitoring deployment steps
+
+=== Developer impact:
+
+Described in "Proposed Change" section
+
+== Implementation:
+
+
+=== Assignee(s):
+
+Primary assignee:
+
+  * anmolbabu<anmolbudugutta@gmail.com>
+  * r0h4n
+
+=== Work Items:
+
+* https://github.com/Tendrl/node-agent/issues/546
+* https://github.com/Tendrl/tendrl-ansible/issues/20
+
+== Dependencies:
+
+None
+
+== Testing:
+
+None
+
+== Documentation impact:
+
+None
+
+== References:
+
+None


anmolbabu:
@r0h4n @nthomas-redhat @shtripat @brainfunked Please review..

@r0h4n I have written this spec from the perspective of changes due to deprecating node-monitoring..
Please add if I am missing anything from the perspective of node-agent, for example:
The best place within node-agent code-base for the items being moved from node-monitoring to reside
Please note, that the plugins that will need to move out of node-monitoring will be majorly related to sds and none or least number of them actually related to node(free from dependencies related to sds).. So, node-monitoring used to package them all in node-monitoring itself they(plugins and templates) only need to be shipped with some package and copied to /usr/lib64/collectd on installation and the flow stuff is generic and can probably reside along with node-agent flows..

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,181 @@
...
+= Refactoring Of Node Monitoring Flows Into Node-Agent
+
+== Introduction
+
+This specification deals with detailing the work involved in merging the
+node-monitoring and node-agent services.
+
+Currently the responsibilities of node-monitoring are as follows:
+
+* It packages the collectd custom plugins
```
shtripat, line 12: I remember once we talked about possibilities of these collectd plugins to gluster org itsefl. May be for future.

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+* Earlier the jobs for intiating the flows to configure collectd on nodes were
+  initiated by performance-monitoring node. But for now(momentarily) as a final
+  step of import, create and expand cluster, this will be intiated on all nodes
+  of cluster.
+
+  ----
+  Note:
+  Initiating configuration on cluster nodes from import cluster flow will
+  involve a post atom that can load a job to etcd on the lines of:
```
shtripat: A post atom creating a job to do the work. Why not node local post atom can take care. Anyway import flow runs on all the gluster nodes?
Yes, it might be an issue for ceph cluster where, import runs on mon nodes only. We might have to look at the case of ceph at later stage may be. @r0h4n ??

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,181 @@
...
+==== Notifications/Monitoring impact:
+
+node-monitoring project will be deprecated.
+
+==== Tendrl/common impact:
+None
+
+==== Tendrl/node_agent impact:
+
+As described in section "Proposed Change" section, collectd templates, custom
```
shtripat, line 130: Can we decide and finalize the drop dir structure for these components here?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: yeah I think the only inhibitor to that is the our namespacing of the metrics which includes cluster-id. But yeah we can always get that as a configurable parameter and then facilitate its use...
But yes we need to finalize on where these plugins will go in other words how its packaged

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Yep waiting on inputs on https://github.com/Tendrl/specifications/pull/219#issuecomment-318278257 to close this part

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
r0h4n: the node-agent namespace will provide the cluster id via NS.tendrl_context.integration_id if that node is part of a cluster. So please move the plugins to node-agent

anmolbabu: @r0h4n Yes for now its going to be as you said..
My comment was to handle the case when we move this gluster space..
Which is when gluster does not know how and what cluster id is which is why the problem then is what happens to cluster-id part of metric namespace.. So, we'll need to keep the prefix part(tendrl.clusters.cluster-id or tendrl.cluster.) of namespace as configurable(only to the extent of handling cases of presence/absence of cluster-id) but may not be for now :) ..

The spec is currently in accordance with the decisions taken for now...
It can evolve later when required..

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,223 @@
...
+plugins and flow to generate collectd configuration will move into node-agent.
+
+* The plugins will go as under :
+
+----
+[node_agent]$ tree tendrl/node_agent/collectors/
+tendrl/node_agent/collectors/
```
r0h4n: tendrl/node_agent/collectd/collectors

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,223 @@
...
+├── tendrl_ceph_utilization.py
+├── tendrl_glusterfs_base.py
+└── tendrl_glusterfs_clients_info.py
+----
+
+* The templates will be accomodated as in:
+
+----
+[node_agent]$ tree tendrl/node_agent/templates/
+tendrl/node_agent/templates/
```
r0h4n: tendrl/node_agent/collectd/templates/

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,223 @@
...
+None
+
+==== Notifications/Monitoring impact:
+
+node-monitoring project will be deprecated.
+
+==== Tendrl/common impact:
+None
+
+==== Tendrl/node_agent impact:
```
r0h4n, line 128: Provide details of triggering "ConfigureCollectd" after success if ImportCluster

r0h4n (changes requested):

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

r0h4n: where?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
r0h4n: where?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Updated PR with changes

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Updated PR with changes

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+packages was to facilitate modularity in tendrl stack such that users of tendrl
+stack can have the freedom to choose whether or not to have monitoring stack
+along with tendrl core stack. But, this pulls in overhead of having to maintain
+greater number of services.So, the idea is to now consolidate the services as
+much as possible and as part of this idea, this spec proposes consolidation of
+node-monitoring into node-agent.
+
+== Use Cases
+
+Admin creates/imports a cluster using/into tendrl and has monitoring enabled
```
cybernth, line 32: Currently we deal with only the import case

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+* A generic flow and a python script(that can be used as a linux command) to
+  get the configs generated from jinja templates given the appropriate parameters.
+
+== Problem description
+
+The idea behind having node-agent and node-monitoring seperated into a seperate
+packages was to facilitate modularity in tendrl stack such that users of tendrl
+stack can have the freedom to choose whether or not to have monitoring stack
+along with tendrl core stack. But, this pulls in overhead of having to maintain
+greater number of services.So, the idea is to now consolidate the services as
```
cybernth, line 26: Basically maintaining a service just for a single flow :)

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+and the plugins currently in:
+
+----
+https://github.com/Tendrl/node-monitoring/tree/master/tendrl/node_monitoring/plugins
+----
+
+will be moved into node-agent.
+
+* Earlier the jobs for intiating the flows to configure collectd on nodes were
```
cybernth: How/when the collectd configuration is loaded on the node where etcd/API is running?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+None
+
+==== Notifications/Monitoring impact:
+
+node-monitoring project will be deprecated.
```
cybernth: what about `performance-monitoring`? will that be covered in a separate spec?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+│   ├── tendrl_glusterfs_peer_network_throughput.py
+│   ├── tendrl_glusterfs_profile_info.py
+│   └── utils.py
+├── ini2json.py
+├── __init__.py
+├── tendrl_ceph_cluster_iops.py
+├── tendrl_ceph_node_network_throughput.py
+├── tendrl_ceph_utilization.py
+├── tendrl_glusterfs_base.py
+└── tendrl_glusterfs_clients_info.py
```
cybernth: Why few of the plugins(gluster) are kept under glusterfs and others are not?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,224 @@
...
+├── latency.jinja
+├── memory.jinja
+├── mount_point.jinja
+├── network.jinja
+├── swap.jinja
+├── tendrl_ceph_cluster_iops.jinja
+├── tendrl_ceph_cluster_utilization.jinja
+├── tendrl_ceph_node_network_throughput.jinja
+├── tendrl_glusterfs_clients_info.jinja
+└── tendrl_glusterfs.jinja
```
cybernth: its better to organize the templates into specific folders like ceph, gluster, common etc

cybernth (changes requested):

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Ack. But the main reason was probably to keep monitoring separate..

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Ack.. Just wanted to keep the spec as much open as possible as it anyway doesn't bring any overhead relevant to this spec :)

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: No a node not part of a cluster is never configured for monitoring..
I think this was decided for the purpose having all stats namespaced under relevant cluster-id..
@anivargi Please confirm

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Yes I think @rishubhjain or @GowthamShanmugam spec would cover that part

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Ack. The only reason was that, the plugins are classified/grouped and a base class loads all plugins within a specific directory at once at fixed interval of time.But then, the problem here is, some plugins cannot be executed as frequently as a few others..

So, as the classification of plugins possibly into heavy-weight(less frequently executed) and light -weight(more frequently executed) is currently not implemented, I did it this way.. I'll change it nowAck. The only reason was that, the plugins are classified/grouped and a base class loads all plugins within a specific directory at once at fixed interval of time.But then, the problem here is, some plugins cannot be executed as frequently as a few others..

So, as the classification of plugins possibly into heavy-weight(less frequently executed) and light -weight(more frequently executed) is currently not implemented, I did it this way.. I'll change it now

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Ack

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
cybernth: Tagging @brainfunked , @r0h4n as well

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
cybernth: @anmolbabu Please link that here

anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
shtripat: Yes. For that reason only we try to configure collected post import cluster success.

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,232 @@
...
+
+and the plugins currently in:
+
+----
+https://github.com/Tendrl/node-monitoring/tree/master/tendrl/node_monitoring/plugins
+----
+
+will be moved into node-agent.
+
+* Earlier the jobs for intiating the flows to configure collectd on nodes were
```
shtripat: s/intiating/initiating/

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,232 @@
...
+  initiated by performance-monitoring node. But for now(momentarily) as a final
+  step of import, create and expand cluster, this will be intiated on all nodes
```
shtripat: s/Intiated/initiated/

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,232 @@
...
+=== Data model impact:
+
+A new flow definition will be added to node-agent under its namespace as below:
+
+----
+namespace.node_agent:
+  flows:
+    ConfigureCollectd:
+      atoms:
+        - tendrl.objects.Node.atoms.Cmd
```
shtripat, line 91: Is this correct value?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,232 @@
...
+│   │   │   ├── __init__.py
+│   │   │   ├── tendrl_glusterfs_brick_utilization.py
+│   │   │   ├── tendrl_glusterfs_health_counters.py
+│   │   │   ├── tendrl_glusterfs_peer_network_throughput.py
+│   │   │   └── tendrl_glusterfs_profile_info.py
+│   │   ├── tendrl_glusterfs_base.py
+│   │   └── utils.py
+│   ├── ini2json.py
+│   ├── __init__.py
+│   ├── tendrl_ceph_cluster_iops.py
```
shtripat: For ceph we need a parallel dir to `glusterfs`

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
```diff
@@ -0,0 +1,232 @@
...
+The end user will now have only node-agent to be deployed on nodes apart from
+sds-integrations.
+
+=== Performance impact:
+
+None
+
+=== Other deployer impact:
+
+Tendrl ansible will now need to remove node-monitoring deployment steps
```
shtripat, line 200: Create a wiki issue and add in references section.

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: any reason why this could be incorrect because I have only copy pasted the definition entry as it used to be nnode-monitoring def file(which used to work during our previous drops) and only refactored node-monitoring=>node-agent

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
shtripat: All I meant was the atom `tendrl.objects.Node.atoms.Cmd`. Is this the atom which needs to be executed for `ConfigureCollectd`?

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Done

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
anmolbabu: Ok I thought some refactoring moved it to commons or something like that..

Yeah it needs to be cmd because we have configuration utility as a command which will be installed in /usr/bin which makes it usable like a linux cli command and the flow just invokes it per node per config

cybernth (approved):
Approving this with the assumption that the nodes which are not part of the the cluster is not required to be monitored(including the tendrl node where API,etcd etc. runs).see https://github.com/Tendrl/specifications/pull/219#discussion_r130338879

specs/refactoring_of_node_monitoring_flows_into_node_agent.adoc
r0h4n: What does tendrl install at /usr/bin?  whats that configuration utility

r0h4n (approved):

Issue #225, opened 2017-08-08
Updated cluster list view in the UI to show all detected (managed and unmanaged) clusters


gnehapk:
Cluster list should display both managed and unmanaged clusters.

PR #209, opened 2017-07-25
Gluster Import Cluster Workflow in UI

gnehapk:
tendrl-bug-id: #183

PR #228, opened 2017-08-08
Updated cluster list view to include all detected (managed or unmanaged) clusters

gnehapk:
tendrl/bug-id: #225

### DIFF
--- a/specs/updated_cluster_list_implementation_with_upadated_API_response
+++ b/specs/updated_cluster_list_implementation_with_upadated_API_response
@@ -1,0 +1,162 @@
+:imagesdir: ./images/
+
+= Tendrl Definitions: Display list of unmanaged gluster clusters discovered by Tendrl
+
+Display list of unmanaged gluster clusters present in Tendrl. The list should have functionality for 
+
+* filtering by name.
+* sorting by name, status, cluster version and managed.
+
+Cluster list will be sorted by name by default.
+
+View design will look like:
+
+image::upadtedClusterList.png[Upadted Cluster List]
+
+== Problem description
+
+Tendrl should show list of unmanaged gluster clusters present in Tendrl along with their details.
+
+The details shown for each cluster in UI is -
+
+* Cluster Name
+* Status
+* Managed State
+* Cluster Version
+* Import button
+* Message to state cluster configurable state for eg. ready to import, cluster misconfigured.
+
+The workflow is as follows:
+
+* API will provide the list of clusters to the Tendrl.
+
+* User can click on expand icon of each cluster record to see the error messages and hosts associated with it.
+
+* In the hosts tab, host fqdn and role will be displayed.
+
+* If Managed state is "No" and cluster is ready to import, Import button will be displayed.
+
+* If Managed state is "No" and cluster is not properly configured, Import button will be disabled.
+
+* Clicking Import button will take to import cluster workflow.
+
+== Use Casess
+
+Users can see the list of unmanaged gluster cluster present in Tendrl. They can also 
+
+* filter by name.
+* sort by name, status, cluster version and managed.
+
+== Proposed change
+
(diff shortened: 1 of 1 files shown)


gnehapk:
@r0h4n @brainfunked @nthomas-redhat @anivargi @cloudbehl @a2batic Please review.

specs/updated_cluster_list_implementation_with_upadated_API_response
```diff
@@ -0,0 +1,141 @@
...
+See 'Impacted Modules' section where all the details are populated.
+
+=== Impacted Modules:
+
+==== Tendrl API impact:
+
+/api/1.0/GetClusterList - fetch the list of clusters
+
+Sample API Response - required.
+
```
gnehapk: @anivargi @r0h4n Please provide the updated response format.

specs/updated_cluster_list_implementation_with_upadated_API_response
```diff
@@ -0,0 +1,141 @@
+// vim: tw=79
```
shtripat: Correct the commit headline. There is a spelling mistake :)

specs/updated_cluster_list_implementation_with_upadated_API_response
```diff
@@ -0,0 +1,141 @@
...
+
+* Status
+* Cluster Name
+* Cluster Version
+
+Required from API - 
+
+* Managed State
+* Message to state cluster configurable for eg. ready to import, cluster misconfigured.
+* Host IP address for displaying it under "Hosts" tab of expanded view.
```
r0h4n: What if the host/node has multiple IP addresses? Which IP address do you want to show

r0h4n (changes requested):
Please check my comments on https://github.com/Tendrl/specifications/pull/209

specs/updated_cluster_list_implementation_with_upadated_API_response
gnehapk: @brainfunked @nthomas-redhat @shtripat Which IP address should be shown here?

shtripat: I feel ideally we should show FQDN here (or the IP to which this fqdn resolves to in case IP only to display)

brainfunked: FQDN serves as the identifier for the node. We should also show the IP. Aren't we detecting the cluster network on the backend; @shtripat? We should be able to show only the IP which is part of the cluster itself, in that case.

brainfunked (changes requested):
Looks like the attached image is for the import cluster view. Shouldn't the images showing the cluster list view be included as well?

gnehapk:
@brainfunked By mistaken, I have added it. Will remove it. Thanks.

shtripat:
>> FQDN serves as the identifier for the node. We should also show the IP. Aren't we detecting the cluster
>> network on the backend; @shtripat? We should be able to show only the IP which is part of the
>> cluster itself, in that case.

Currently there is no way to figure out the IP which is used for cluster communication while import cluster. It good to show the FQDN I feel.

specs/updated_cluster_list_implementation_with_upadated_API_response
```diff
@@ -0,0 +1,140 @@
...
+Tendrl should show list of unmanaged gluster clusters present in Tendrl along with their details.
+
+The details shown for each cluster in UI is -
+
+* Cluster Name
+* Status
+* Managed State
+* Cluster Version
+* Import button
+* Message to state cluster configurable state for eg. ready to import, cluster misconfigured.
```
gnehapk: @r0h4n @anivargi Which API will provide the error messages? Can you please provide the sample response of the same?

Issue #224, opened 2017-08-08
Dashboard Spec - Volume Dashboard


julienlim:
# Dashboard Spec - Volume Dashboard

Display a default dashboard for a Gluster volume present in Tendrl that provides at-a-glance information about a single Gluster volumethat includes health and status information, key performance indicators (e.g. IOPS, throughput, etc.), and alerts that can highlight the Tendrl user's (e.g. Gluster Administrator) attention to potential issues in the volume, brick, and disk.

## Problem description

A Gluster Administrator wants to be able to answer the following questions by looking at the cluster dashboard:

* Is my volume up and running, is it healthy?
* Is there a problem with my volume? 
* What’s actually wrong with my storage volume, why it it slow? 
* Is my storage volume filling up too fast?
* When will my storage volume run out of capacity?
* If something is down / broken / failed (e.g. brick down, disk failure, etc.), where and what is the issue, and when did it happen?
* Have the number of clients (indicated via connections) increased (which may possibly be the reason for the performance degradation that the clients / applications are observing?

## Use Cases

Uses Cases in the form of user stories:
* As a Gluster Administrator, I want to view at-a-glance information about my Gluster volume that includes health and status information, key performance indicators (e.g. IOPS, throughput, etc.), and alerts that can highlight my attention to potential issues in the volume, brick, and disk.

* As a Gluster Administrator, I want to compare 1 or more metrics (e.g. IOPS, CPU, Memory, Network Load) across bricks within the volume

* Compare utilization (e.g. IOPS, capacity, etc.) across bricks within a volume

* Look at performance by brick (within a volume) to address diagnosing of RAID 6 disk failure/rebuild/degradation poor performance on one brick

## Proposed change

Provide a pre-canned, default volume dashboard in Grafana (that is initially launchable from the Tendrl UI, and eventually embed it into the Tendrl UI) that shows the following metrics rendered either in text or in a chart/graph depending on the type of metric being displayed below:

The Dashboard is composed of individual Panels (dashboard widgits) arranged on a number of Rows. 

Note: The cluster and volume name or unique identifier should be visible at all times, and user should be able to switch to another volume.

### Row 1

Panel 1: Health
  * show volume status, i.e. Up, Down, Up (Degraded), Up (Partial) -- see https://github.com/gluster/gstatus for details about the various status
  * the color of panel should be green when OK, red when down, yellow when up( degraded), or up (partial)
[...]

PR #229, opened 2017-08-08
Gluster Metrics-MileStone 2

anmolbabu:
tendrl-bug-id: Tendrl/specifications#188

### DIFF
--- a/specs/gluster-metrics-milestone2.adoc
+++ b/specs/gluster-metrics-milestone2.adoc
@@ -1,0 +1,242 @@
+// vim: tw=79
+
+= Gluster Metrics
+
+== Introduction
+
+This specification deals with monitoring tendrl managed gluster cluster.
+It explains the following:
+
+* Which gluster metrics are to be enabled.
+* How each of the metrics is going to be implemented.
+* Any flows to be added to the node agent for configuring any collectd plugins
+  needed by the individual metrics.
+This specification will deal with metrics implemented in milestone 2
+
+== Problem description
+
+Admin using tendrl for gluster cluster management, should be able to see an
+easy to interpret representation of his gluster cluster/s with appropriate
+utilization trend patterns and any issue in his gluster cluster/s should be
+with no/minimal effort be made noticeable to the admin.
+
+== Use Cases
+
+A gluster cluster is imported(or created) in tendrl. From this point, this
+cluster needs to be monitored for utilization and status behaviour.
+
+== Proposed change
+
+The gluster metrics will be enabled at the very level from collectd and any
+aggregations thereon will be done by the grafana either on its own or using the
+aggregation rules as communicated to it by the monitoring-integration module.
+
+* Following are the gluster metrics to be enabled from collectd along with the
+  source of the information:
+    ** Brick level:
+        *** FOP Latency -- gluster volume profile <vol_name> info --xml
+
+            xml attribute hierarchy in cli response:
+            cliOutput -> volProfile -> brick -> cumulativeStats -> fopStats -> fop -> avgLatency
+            cliOutput -> volProfile -> brick -> cumulativeStats -> fopStats -> fop -> minLatency
+            cliOutput -> volProfile -> brick -> cumulativeStats -> fopStats -> fop -> maxLatency
+
+            The grafana aggregates(sum) fop level latencies(avgLatency) into
+            brick level latencies.
+
+        *** LVM Thin Pool metadata usage -- lvm vgs command
+        *** LVM thin pool data usage -- lvm vgs command
+        *** IOPS
+            **** Collectd pushes fop level brick wise fop hits
+                 xml attribute hierarchy in cli response:
+                 cliOutput -> volProfile -> brick -> intervalStats -> fopStats -> fop -> hits
+            **** Grafana aggregates(sum) brick level fop stats
+        *** Throughput
+            **** Collectd pushes totalRead and totalWrite at brick level from
+                 gluster volume profile command o/p
+                 xml attribute hierarchy in cli response:
+                 cliOutput -> volProfile -> brick -> intervalStats -> totalRead
+                 cliOutput -> volProfile -> brick -> intervalStats -> totalWrite
+            **** Grafana aggregates the same as totalRead + totalWrite for iops
+        ----
+        l = map(lambda x: dict(x),
+                map(lambda x: [e.split('=') for e in x],
+                    map(lambda x: x.strip().split('$'), out)))
+        d = {}
+        for i in l:
+            if i['LVM2_LV_ATTR'][0] == 't':
+                k = "%s/%s" % (i['LVM2_VG_NAME'], i['LVM2_LV_NAME'])
+            else:
+                k = os.path.realpath(i['LVM2_LV_PATH'])
+            d.update({k: i})
+        return d
+
+        if lvs and device in lvs and \
+           lvs[device]['LVM2_LV_ATTR'][0] == 'V':
+            thinpool = "%s/%s" % (lvs[device]['LVM2_VG_NAME'],
+                                  lvs[device]['LVM2_POOL_LV'])
+            out['thinpool_size'] = float(
+                lvs[thinpool]['LVM2_LV_SIZE']) / 1024
+            out['thinpool_used_percent'] = float(
+                lvs[thinpool]['LVM2_DATA_PERCENT'])
+            out['metadata_size'] = float(
+                lvs[thinpool]['LVM2_LV_METADATA_SIZE']) / 1024
+            out['metadata_used_percent'] = float(
+                lvs[thinpool]['LVM2_METADATA_PERCENT'])
+            out['thinpool_free'] = out['thinpool_size'] * (
+                1 - out['thinpool_used_percent'] / 100.0)
+            out['thinpool_used'] = out['thinpool_size'] - out['thinpool_free']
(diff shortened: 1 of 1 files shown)


anmolbabu:
This is a WIP PR

@nthomas-redhat @shtripat @r0h4n @brainfunked Please review

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+            out['metadata_used'] = out['metadata_size'] - out['metadata_free']
+        ----
+
+    ** Volume level:
+        *** Pending Heal -- gluster volume heal all statistics
+        *** Total Heal -- gluster volume heal all statistics
+        *** Heal pending Split Brain files
+        *** Rebalance status -- gluster get-state glusterd..
+        *** Geo-rep sessions count
+            **** Faulty  -- gluster get-state -- slated for upcoming glusterfs builds
```
shtripat: This is available now in nightly master builds.

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+            **** Ok      -- gluster get-state -- slated for upcoming glusterfs builds
+        *** Number of connections -- gluster get-state
+        *** Snapshot count -- gluster get-state
+        *** Volume status(Degraded)
```
shtripat: Based on calculations the volume states could be one from `up`, `down`, `up (partial)` and `up (degraded)`

anmolbabu: How this is calculated??

Based on cosntituent brick statuses??
Then it can't be done here..
Because collectd plugin sits on end nodes and end node get-state o/p contsins brick statuses of only that node..

This will move into monitoring-integration then

shtripat: Yes, it should be done as part of monitoring-integration. The logic can be referred from https://github.com/Tendrl/gluster-integration/pull/347 which decides the volume's state based on individual participating bricks and type volume type.

Make a note this is different from volume status which has only two values `Started` and `Stopped`.

shtripat:
Based on discussion with Aravinda, the volume level geo-rep session could be in one of the states below
* up
* partial
* down

The logic to calculate these values could be as below

```
if no of non-faulty pairs is zero (all the pairs status is faulty):
    volume geo-rep status = down
else:
    if no of non faulty pairs is equal to the no of bricks in the volume (all the pairs are either active/passive):
        volume geo-rep status = up
    else:
        volume geo-rep status = partial
```

shtripat:
As per latest discussion with @nthomas-redhat amd @nnDarshan we would add volume level aggregated geo-rep status value as per above logic. Basically it would be a field like `cluster/{int-id}/Volumes/{vol-id}/GeoReplicationSessions/{session_id}/status` which would be an aggregated value from the geo-rep pair status values.

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+== Proposed change
+
+The gluster metrics will be enabled at the very level from collectd and any
+aggregations thereon will be done by the grafana either on its own or using the
+aggregation rules as communicated to it by the monitoring-integration module.
+
+* Following are the gluster metrics to be enabled from collectd along with the
+  source of the information:
+    ** Brick level:
+        *** FOP Latency -- gluster volume profile <vol_name> info --xml
```
cybernth, line 37: We need iops, latency and throughput at multiple levels(brick, volume and cluster). What are the stats available and how the computation is done?

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+                lvs[thinpool]['LVM2_DATA_PERCENT'])
+            out['metadata_size'] = float(
+                lvs[thinpool]['LVM2_LV_METADATA_SIZE']) / 1024
+            out['metadata_used_percent'] = float(
+                lvs[thinpool]['LVM2_METADATA_PERCENT'])
+            out['thinpool_free'] = out['thinpool_size'] * (
+                1 - out['thinpool_used_percent'] / 100.0)
+            out['thinpool_used'] = out['thinpool_size'] - out['thinpool_free']
+            out['metadata_free'] = out['metadata_size'] * (
+                1 - out['metadata_used_percent'] / 100.0)
```
cybernth, line 90: With respect to stats belongs to bricks, I assume below are implemented:
Brick status
Connections
Capacity Utilization
Capacity Available

What's plan on below(not implemented aand not part of this spec):
 Inodes Utilization
Inodes Available

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+            out['metadata_used'] = out['metadata_size'] - out['metadata_free']
+        ----
+
+    ** Volume level:
+        *** Pending Heal -- gluster volume heal all statistics
+        *** Total Heal -- gluster volume heal all statistics
+        *** Heal pending Split Brain files
```
cybernth: How map these parameters against healinfo output i.e.
No. of entries healed
No. of entries in split-brain
No. of heal failed entries
Specifically Pending Heal and Total Heal

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+        *** Rebalance status -- gluster get-state glusterd..
```
cybernth: n in progress and n failures are required. Please add this to doc , also which command is used and computation if any

specs/gluster-metrics-milestone2.adoc
cybernth: Add the details to the spec. Also have a look at https://github.com/Tendrl/specifications/issues/224#issuecomment-321579201
Require which command is used and computation if any

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+        *** Geo-rep sessions count
+            **** Faulty  -- gluster get-state -- slated for upcoming glusterfs builds
+            **** Ok      -- gluster get-state -- slated for upcoming glusterfs builds
+        *** Number of connections -- gluster get-state
+        *** Snapshot count -- gluster get-state
+        *** Volume status(Degraded)
+    ** Cluster level:
+        *** Quorum status
```
cybernth: Add details to the spec about commands used and computation

specs/gluster-metrics-milestone2.adoc
```diff
@@ -0,0 +1,229 @@
...
+        *** Degraded volume count
+    ** Brick and cluster status wise counters can't be made  available from the
+       level of collectd as only the brick info corresponding to node on which
+       the collectd plugin is currently in execution, will be accessible and
+       info regarding the other bricks in cluster is not available and plugins
```
cybernth: Link the details to this spec

cybernth (changes requested):

specs/gluster-metrics-milestone2.adoc
anmolbabu: Part of other spec https://github.com/anmolbabu/specifications/blob/e27841f74a59fbbc859359f0e6819b4b7a3b77ef/specs/gluster_metrics-milestone1.adoc

specs/gluster-metrics-milestone2.adoc
anmolbabu: Removing this as they have been implemented as part of monitoring-integration
Cc: @cloudbehl @rishubhjain

cloudbehl: @anmolbabu removing Volume status(Degraded) or anything else also?

specs/gluster-metrics-milestone2.adoc
anmolbabu: Removing this as it has been added by monitoring-integration

Cc @cloudbehl @rishubhjain

specs/gluster-metrics-milestone2.adoc
anmolbabu: Ack

specs/gluster-metrics-milestone2.adoc
anmolbabu: Added the details in spec
Reference: https://github.com/Tendrl/specifications/issues/224#issuecomment-321579201 and details here

specs/gluster-metrics-milestone2.adoc
anmolbabu: No computation involved adding the command details

specs/gluster-metrics-milestone2.adoc
anmolbabu: Ack

specs/gluster-metrics-milestone2.adoc
anmolbabu: Sorry I didn't understand Could you please elaborate?
I have added the details in this spec itself

specs/gluster-metrics-milestone2.adoc
anmolbabu: Quorum status also

anmolbabu: Quorum status
Volume status wise counts
Geo-rep related status wise counters
@cloudbehl

specs/gluster-metrics-milestone2.adoc
anmolbabu: Cc: @cloudbehl

Issue #216, opened 2017-07-26
Gluster Trusted Storage Pool details view


gnehapk:

brainfunked:
Details for individual Gluster storage pools that include essential health information, list of all the volumes, hosts and bricks participating in the pool, mapping of volumes to bricks and bricks to hosts etc.

brainfunked:
Point in time utlisation of hosts' CPU, memory and storage utlisation is not part of this spec. There are two corresponding issues at https://github.com/Tendrl/ui/issues/513 and https://github.com/Tendrl/node-agent/issues/585 where this is being tracked.

PR #232, opened 2017-08-09
Specification for gluster object details view

gnehapk:
tendrl-bug-id: #216


gnehapk:
@r0h4n @brainfunked @nthomas-redhat @anivargi @shtripat @cloudbehl @a2batic Please review.

specs/gluster_object_detail_views.adoc
```diff
@@ -0,0 +1,198 @@
...
+The details shown for each host are -
+
+* Status
+* Name
+* Role
+* Alerts
+
+The host list should have functionality for 
+
+* filtering by name, role and status.
```
shtripat, line 20: In case of gluster all the nodes have same role. But have he filtering mechanism in place which would be helpful later.

specs/gluster_object_detail_views.adoc
```diff
@@ -0,0 +1,198 @@
...
+image::hostList.png[Gluster specific host list view]
+
+*Volume List*
+
+The details shown for each volume are -
+
+* Status
+* Name
+* Type
+* Bricks
```
shtripat: This is just a bricks count or something `10 (8 up, 2 down)` ?
The screen design says value `Server` for brick. Do you mean it would list all the brick names with full path in this list?

specs/gluster_object_detail_views.adoc
```diff
@@ -0,0 +1,198 @@
...
+
+== Use Casess
+
+Display list of volume and host associated with a gluster cluster present in Tendrl. 
+
+== Proposed change
+
+* Create a view for cluster object detail which will have two tab: Hosts and Volumes.
+* Create components for hosts and volumes list page.
+* Implement client side filer and sorting functionality using angular's filter and orderby components.
```
shtripat: s/filer/filter/

specs/gluster_object_detail_views.adoc
```diff
@@ -0,0 +1,198 @@
...
+* Send polling call after 10 seconds for volume list and 30 seconds for host list to get the updated data.
```
shtripat: 30 sec polling for hosts of a cluster which is already imported might not be required as this data doesnt change frequently. This could a minute or so. @Tendrl/tendrl-core comments?

specs/gluster_object_detail_views.adoc
gnehapk: @shtripat Can you please specify which filtering mechanism you are referring here.

specs/gluster_object_detail_views.adoc
gnehapk: @shtripat Yes, its a brick count. Will update the mockup. Thanks.

specs/gluster_object_detail_views.adoc
shtripat: I meant filter by role here.

specs/gluster_object_detail_views.adoc
shtripat: Ack

specs/gluster_object_detail_views.adoc
shtripat: So in case of gluster all the node are `gluster/server` whereas in ceph there are two roles applicable for nodes namely `ceph/mon` and `ceph/osd`. At the moment you can have only one role `gluster/server` defined for role based filtering.

brainfunked: If the same host list view would be used for other storage systems such as Ceph, then the `role` column is essential. Even in gluster's case, only a specific node would have the role of provisioner (cluster specific and/or global), which needs to be shown.

The UI would need to add specific logic to be able to display the provisioner roles. I figure this logic would be to look for the tags `provisioner/gluster` and `provisioner/<cluster_id>`. These roles would map to the terms `Global Gluster Provisioner` and `Provisioner` respectively.

BTW, should the role `gluster/server` be renamed to `gluster/peer` on the backend? I believe that's the official gluster terminology. When changed, this would also enable the role for every node in gluster to be displayed as `Peer` in the UI.

@gnehapk @shtripat @nnDarshan @r0h4n

brainfunked (changes requested):
@shtripat @nnDarshan does the backend present a straightforward way of figuring out the number of bricks on a host in the cluster and the number of volumes the host participates in?

shtripat:
>> should the role gluster/server be renamed to gluster/peer

I am fine with this requirement

gnehapk:
@nthomas-redhat @brainfunked @shtripat @r0h4n I have updated the spec. Please review.

specs/gluster_object_detail_views.adoc
shtripat: I am ok with `gluster/peer`

Issue #176, opened 2017-07-19
User Management in the Tendrl UI


gnehapk:
Allow views and CRUD operations based on the roles to all users.
This spec focuses on the tendrl ui and the corresponding api changes.

anivargi:
@gnehapk associate the spec to the issue.

PR #235, opened 2017-08-10
#176 User Management: Adding Users

a2batic:
Bug-id: #176
Signed-off-by: a2batic <murarkakanika@gmail.com>

### DIFF
--- a/specs/Spec_for_adding_user.adoc
+++ b/specs/Spec_for_adding_user.adoc
@@ -1,0 +1,171 @@
+# User Management: Adding new User in Tendrl
+
+## Problem description
+
+This spec talks about Adding/Creating a new User to Tendrl by the Admin
+
+## Use Cases
+
+The flow goes as follows:-
+
+* A user logs in as Admin
+* Clicks on Admin tab in Navigation
+* The navigation drops down
+* Clicks on *Users*
+* Clicks on *Add* button in Users page view
+
+User will land onto https://redhat.invisionapp.com/share/KNB25OEQT#/screens/226063802[Add Users] page
+
+## Proposed change
+
+### UI changes
+
+* Add *Users* tab in Admin navigation tab
+* Add *Add User* Button in Users Page view
+* Add a form for adding a user. The form will contain following feilds:-
+** User ID
+*** Input type : text
+*** Validations:
+**** Should be Unique
+**** Minimum length: 4 characters
+**** Should be mandatory
+** First Name
+*** Input type : text
+*** Validations:
+**** Should be mandatory
+** Last Name
+*** Input type : text
+** Email
+*** Input type : email
+*** Validations:
+**** Should be unique
+**** Validate email ID
+**** Should be mandatory
+** Email Notification
+*** Input type : checkbox
+*** Default : Enabled
+** Password
+*** Input type : password
+*** Validations:
+**** Minimum length: 8 characters
+**** Should be mandatory
+** Confirm Password
+*** Input type : password
+*** Validations:
+**** Should be mandatory
+**** Should match with the password
+** Set a Role
+*** Admin
+*** Normal user
+*** Limited user
+*** Input type : Radio button
+*** Default : admin
+
+*NOTE*: 
+
+* The fields ```First Name``` and ```Last Name``` will be assigned value from the API property ```name``` by splitting it at space.
+* Maximum length for field ```name``` : 100 Characters
+
+### Impacted Modules:
+
+#### Tendrl API impact:
+
+```
+POST /api/1.0/users - creates a new user
+```
+
+* Current sample API response: 
+
+```
+{
+	"email": "thardy@tendrl.org",
+	"username": "thardy",
+	"name": "Tom Hardy",
+	"role": "admin"
+}
+
+```
+
+* Sample API response Required:
+
+```
+{
+	"username": "JDoe1",
+	"name" : "John Doe",
+	"email": "jdoe@tendrl.org",
+	"email_notifications": "true",
+	"role": "admin"
+}
+
+```
+
+API support present for -
+
+* email
+* username
+* name
+* role
+
+Required from API - 
+
+* email_notifications
+
+#### Notifications/Monitoring impact:
+
+If ```email_notifications = true ```, email notifications are required to be monitored 
+
+### Security impact:
+
+### Other end user impact:
+
+None
+
+### Performance impact:
+
+None
+
+### Other deployer impact:
+
+None
+
+### Developer impact:
+
+None
+
+## Implementation:
+
+
+### Assignee(s):
+
+Primary assignee:
+  gnehapk
+
+Other contributors:
+  a2batic
+  
+### Work Items:
+
+https://github.com/Tendrl/specifications/issues/176
+
+
+## Dependencies:
+
+Need API support for required additional fields:
+
+* email_notifications
+
+
+## Testing:
+
+* Test whether fields of adding user form are validating.
+* Check if user is actually getting created and displayed on user list view with correct details.
+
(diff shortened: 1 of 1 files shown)


a2batic:
@anivargi @brainfunked @cloudbehl @julienlim @gnehapk, please review

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+## Problem description
+
+This spec talks about Adding/Creating a new User to Tendrl by the Admin
+
+## Use Cases
+
+The flow goes as follows:-
+
+* A user login as Admin
```
shtripat: s/login/logs in/

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+### UI changes
+
+* Add *Users* tab in Admin navigation tab
+* Add *Add User* Button in Users Page view
+* Add a form for adding a user. The form will contain following feilds:-
+** User ID
+*** Input type : text
+*** Validations:
+**** Should not contain special characters
```
shtripat: special chars like `_`, `-`, '.' should be allowed? Should start with alphanumeric char.

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+**** Should be Unique
+**** Can be of a particular Maximum length
```
shtripat: There should be a min length check as well

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+```
+POST /api/1.0/users - creates a new user
+```
+
+* Sample API response now : 
+
+```
+{
+	"email": "thardy@tendrl.org",
```
shtripat: Where is first name and last name here? Is it like both merged in `name` field?

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+Need API support for required additional fields:
+
+* first_name
+* last_name
+* email_notifications
+
+
+## Testing:
+
+Test whether fields of adding user form are validating.
```
shtripat: Check if user is actually getting created and they are able to login to the system and UI is displayed as per their role :)

specs/Spec_for_adding_user.adoc
a2batic: Acknowledge

specs/Spec_for_adding_user.adoc
a2batic: Acknowledge, I will elaborate it more

specs/Spec_for_adding_user.adoc
a2batic: Yes there is a minimum length too, I will update it

specs/Spec_for_adding_user.adoc
a2batic: I will add this :)

specs/Spec_for_adding_user.adoc
a2batic: Initially, we just had ```name``` field, now according to new mockup https://redhat.invisionapp.com/share/KNB25OEQT#/screens/226063802, we require separate first name and last name

a2batic: Or Should we modify the ```name``` only by using JavaScript. What do you suggest @anivargi @gnehapk ?

specs/Spec_for_adding_user.adoc
gnehapk: @brainfunked @anivargi @nthomas-redhat What validations are required for User ID field? If we want to check for uniqueness then we will require an API call to do that. What should be the minimum and maximum length allowed for this field.

specs/Spec_for_adding_user.adoc
gnehapk: @a2batic If it is possible for API to provide it, then we can have both fields in UI.  @anivargi Is it doable ?

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,166 @@
...
+* username
+* name
+* role
+
+Required from API - 
+
+* first_name
+* last_name
+* email_notifications
+
```
gnehapk: @anivargi Please provide the API response for these fields.

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+* email
+* username
+* name
+* role
+
+Required from API - 
+
+* email_notifications
+
```
a2batic, line 112: @anivargi, please review

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+** Confirm Password
+*** Input type : password
+*** Validations:
+**** Should be mandatory
+** Set a Role
+*** Admin
+*** Normal user
+*** Limited user
+*** Input type : Radio button
+
```
cybernth: Are we not suporting `Read-only` user?

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+*** Validations:
+**** Should be mandatory
+** Last Name
+*** Input type : text
+*** Validations:
+**** Should be mandatory
+** Email
+*** Input type : email
+*** Validations:
+**** Validate email ID
```
cybernth: Should this be unique as well?

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+## Proposed change
+
+### UI changes
+
+* Add *Users* tab in Admin navigation tab
+* Add *Add User* Button in Users Page view
+* Add a form for adding a user. The form will contain following feilds:-
+** User ID
+*** Input type : text
+*** Validations:
```
cybernth, line 28: Where all the validations are done? UI or API?

specs/Spec_for_adding_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+**** Should be mandatory
+** Set a Role
+*** Admin
+*** Normal user
+*** Limited user
+*** Input type : Radio button
+
+*NOTE*: 
+
```
cybernth: I think we are not going to support enable/disable user. If yes we need to call out explicitly as it is there is in the design

cybernth (changes requested):

specs/Spec_for_adding_user.adoc
a2batic: All validations are done on API only

specs/Spec_for_adding_user.adoc
a2batic: @anivargi, Is there any such validation for email field ?

specs/Spec_for_adding_user.adoc
a2batic: @nthomas-redhat, According to https://github.com/Tendrl/api/blob/master/docs/authorizaton.adoc , Limited user is same as Read-only user

specs/Spec_for_adding_user.adoc
a2batic: So, should we comment it out from UI for now ?

a2batic: @nthomas-redhat, Thanks for reviewing !!

specs/Spec_for_adding_user.adoc
anivargi: Username and email are unique, checked by the API.

PR #238, opened 2017-08-11
#176 User Management: Editing User

a2batic:
Bug-id: #176
Signed-off-by: a2batic <murarkakanika@gmail.com>

### DIFF
--- a/specs/Spec_for_editing_user.adoc
+++ b/specs/Spec_for_editing_user.adoc
@@ -1,0 +1,167 @@
+# User Management: Editing User view for Admin
+
+## Problem description
+
+This spec talks about viewing or editing information stored about a single user.
+
+## Use Cases
+
+The flow goes as follows:-
+
+* A user logs in as Admin
+* Clicks on Admin tab in Navigation
+* The navigation drops down
+* Clicks on *Users*
+* Clicks on *Edit* option for a particular user
+
+User will land onto https://redhat.invisionapp.com/share/KNB25OEQT#/screens/226063805[Edit User] view page
+
+## Proposed workflow
+
+### UI
+
+Edit user view consist of a form:-
+
+* ```User ID``` - Not Editable
+* First Name
+** Input type: text
+* Last Name
+** Input type: text
+* Email
+** Input type: email
+** validation : Input should be a valid email
+* Email Notifications - whether user should receive notifications
+** Input type: checkbox
+* Password
+** Input type: password
+** Validation: Should be minimum of 8 characters
+** Should contain a Show/Hide button for password visibility. The default visibility will be *hidden*.
+* Confirm Password
+** Input type: password
+** Validation: Should match password
+** Should contain a Show/Hide button for password visibility. The default visibility will be *hidden*.
+* Select a Role
+** Roles: Admin, Normal, Limited
+** The default will be the current assigned role of User
+* ```Save``` button
+** This button commits changes and returns to User List view page.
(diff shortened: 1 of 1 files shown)


a2batic:
@anivargi @shtripat @nthomas-redhat @brainfunked @gnehapk @julienlim @cloudbehl , please review

specs/Spec_for_editing_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+User will land onto https://redhat.invisionapp.com/share/KNB25OEQT#/screens/226063805[Edit User] view page
+
+## Proposed workflow
+
+### UI
+
+Edit user view consist of a form:-
+
+* ```User ID``` - Not Editable
+* ```Type : Internal``` - Not Editable
```
shtripat: What is type internal? Is it like `admin` is internal user and cannot be edited?

specs/Spec_for_editing_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+* Password
+** Input type: password
+** Validation: Should be minimum of 8 characters
+** Should contain a Show/Hide button for password visibility. The default visibility will be *hidden*.
+* Select a Role
+** Roles: Admin, Normal, Limited
+** The default will be the current assigned role of User
+* ```Save``` button
+** Enabled when User makes change in this form.
+** This button commits changes and returns to User List view page.
```
shtripat: We should show a messages as well that user edit was successful as a balloon I feel. Do we have that?

specs/Spec_for_editing_user.adoc
a2batic: It is related to LDAP, so has to be removed, I will update it

specs/Spec_for_editing_user.adoc
a2batic: @gnehapk, @brainfunked

a2batic: @julienlim , @mcarrano, Can you please suggest if there should be a pop up or notification for this?

specs/Spec_for_editing_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+API support present for -
+
+* email
+* username
+* name
+* role
+
+Required from API - 
+
+* first_name
```
cybernth: Per https://github.com/Tendrl/specifications/pull/235/commits/819f0a8b7323a6c72ca37d7864cfea6973a8583a#diff-047658fc19cc3b6a6cba0d6749752529R65,  you need to concatenate first name and last name and send across to API?

specs/Spec_for_editing_user.adoc
```diff
@@ -0,0 +1,170 @@
...
+* last_name
+* status
```
cybernth: are we going to support the status?

cybernth (changes requested):

specs/Spec_for_editing_user.adoc
a2batic: Yes, I have to update that

specs/Spec_for_editing_user.adoc
a2batic: @anivargi ^^

specs/Spec_for_editing_user.adoc
julienlim: @a2batic No pop-up or notification needed.

